PR #17, opened 2025-05-22
Add schema support

atilafassina:
This PR adds support to SQL seed scripts

---

## To test it

1. pull this branch, go to `/examples/tanstack-start`
2. observe `schema.sql` on root of this app, adjust the schema to your needs
3. run the dev server

Easiest way to confirm the schema and data have been pushed is to claim the newly created DB to your account and check on Neon's Console.


packages/vite-plugin-postgres/src/index.ts
```diff
@@ -6,21 +6,27 @@ import { loadEnv, type Plugin as VitePlugin } from "vite";
...
+	},
 };
 
 let claimProcessStarted = false;
 
 export default function postgresPlugin(
 	options?: Partial<PostgresPluginOptions>,
 ): VitePlugin {
-	const { env: envPath, envKey, referrer } = { ...DEFAULTS, ...options };
+	const { env: envPath, envKey } = { ...DEFAULTS, ...options };
```
guillaumervls, line 29: No sure what's the reason to remove this parameter. We'll want to be able to distinguish between the different projects/starters that'll use this Vite plugin (like Tanstack Start...). I'd keep it here.

atilafassina: that's a user-facing config. So it's less likely that we can we rely on.

E.g.: `@neondatabase:vite-plugin:my-app` instead of the integration layer.
I'll add it back - but I think this can fragment vite-plugin metrics and make them harder to read.

packages/neondb/package.json
```diff
@@ -23,6 +23,9 @@
 		"neondb": "dist/cli.js"
 	},
 	"exports": {
+		"./json-schema": {
```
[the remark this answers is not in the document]
atilafassina: On the previous implementation, when offering a JSON interface, we were generating a schema exporting it so users could have intelissense by adding the `$schema` field. It doesn't make sense anymore, I'll remove it.

packages/vite-plugin-postgres/src/index.ts
guillaumervls: Yep but the same is true already with the `instantNeon` function. _Some_ people can always pass 💩 there. But if/when we do another starter/framework than Tanstack, we'll be happy to be able to pass a parameter. And we can always query our analytics with "starts with" queries, not "exact equals".

packages/vite-plugin-postgres/README.md
```diff
@@ -30,31 +30,48 @@ export default defineConfig({
...
 
-| Option   | Type   | Description                      | Default        |
-| -------- | ------ | -------------------------------- | -------------- |
-| `env`    | string | Path to the `.env` file          | `.env`         |
-| `envKey` | string | Name of the environment variable | `DATABASE_URL` |
+| Option     | Type   | Description                            | Default        |
+| ---------- | ------ | -------------------------------------- | -------------- |
+| `env`      | string | Path to the `.env` file                | `.env`         |
+| `envKey`   | string | Name of the environment variable       | `DATABASE_URL` |
+| `onCreate` | object | Configuration for seeding the database | -              |
```
guillaumervls, line 50: I think you mean `seed` here?

packages/vite-plugin-postgres/README.md
```diff
@@ -30,31 +30,48 @@ export default defineConfig({
...
+| Property | Type   | Description                                     | Default |
+| -------- | ------ | ----------------------------------------------- | ------- |
+| `type`   | string | Type of seeding (currently only `"sql-script"`) | -       |
+| `path`   | string | Path to SQL file to execute after creation      | -       |
 
 ## What gets written
 
 -   The plugin writes both a direct connection string and a pooled connection string to your `.env`.
 -   It also provides a claim URL (valid for 7 days) to take ownership of the database.
+-   If `onCreate` is configured, the specified SQL script will be executed after database creation.
```
guillaumervls, line 63: (`seed` as well)

packages/vite-plugin-postgres/README.md
```diff
@@ -30,31 +30,48 @@ export default defineConfig({
...
 
 ## Type Definitions
 
 ```ts
 interface PostgresPluginOptions {
 	env: string; // Path to the .env file
 	envKey: string; // Name of the environment variable
+	onCreate?: {
```
guillaumervls, line 71: (seed as well)

.changeset/quick-loops-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+"@neondatabase/vite-plugin-postgres": minor
+---
+
+Add `onCreate` option to seed database with SQL script on initialization. This option accepts a path to a SQL file that will be executed after the database is created.
```
guillaumervls, line 5: I think you mean `seed` here?

examples/tanstack-start/app.config.ts
```diff
@@ -4,7 +4,12 @@ import tsConfigPaths from "vite-tsconfig-paths";
 export default defineConfig({
 	vite: {
 		plugins: [
-			postgresPlugin(),
+			postgresPlugin({
+				seed: {
+					type: "sql-script",
+					path: "./schema.sql",
+				},
```
guillaumervls, line 11: If people copy-paste this the `referrer` would be unknown.
Let's put a value like `github:neondatabase/neondb-cli/examples/tanstack-start/app.config.ts`, and add a comment telling them they should change this ASAP with their project's name:
```js
// IMPORTANT: This helps us understand where DB creations on Neon come from.
// Please change this as soon as you can, to add your project name:
// "github:org/your-repo" or "npm:your-package"
```

guillaumervls (commented):
I see you removed `schema.json`, probably those lines can be simplified in the release workflow file https://github.com/neondatabase/neondb-cli/blob/c1bf20c61bd64f35a071f74e95efadca56dd2765/.github/workflows/release.yml#L80-L94

packages/neondb/src/lib/seed-database.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export async function seedDatabase(
+	schemaFilePath: string,
+	connectionString: string,
+) {
+	const client = neon(connectionString);
+	const commands = getSqlCommands(schemaFilePath);
+
+	for (const command of commands) {
+		await client.query(command);
```
guillaumervls, line 12: [curious/not sure] Have you tried just passing the whole file? Do we really need to spilt the commands?

packages/neondb/src/lib/utils/format.ts
```diff
@@ -1,3 +1,5 @@
+import { log } from "@clack/prompts";
```
guillaumervls, line 1: This is not needed, right?

packages/neondb/src/lib/seed-database.ts
guillaumervls: 🙏 thanks for the context! Yeah your solution is perfectly fine.

packages/neondb/src/lib/utils/format.ts
atilafassina: No 🙈 

ffs, I think I'm going to add [knip](https://knip.dev) after this PR

PR #39, opened 2025-06-20
Switch to ASCII Box Art Neon Logo

ruf-io:
## Overview

I spoke briefly with @atilafassina about contributing this very important improvement of an updated Neon logo in ASCII box art instead of the auto-generated characters...

Before:
image

After:
image

I also made the Neon URL absolute so that it was clickable in certain terminals like VS Code


atilafassina (approved):
Right on time for the next release!! 🏆

Thanks!

PR #180, opened 2026-06-10
fix: add ability to pass API host to use

aluuu:
## Why?

`neonctl config status` / `config plan` / `config apply` / `deploy`, `neonctl env`, `neon dev`, and the checkout/link env-pull always talk to the **hardcoded production API host**, even when the user points neonctl at another environment via `NEON_API_HOST` or `--api-host`. Against staging this fails with:

```
ERROR: getProject(...) failed: the Bearer token sent to the Neon API was rejected.
Neon API said: "supplied credentials do not pass authentication"
```

The root cause: every runtime entry funnels into `createNeonApiFromOptions` (`packages/config/src/lib/auth.ts`), which resolves the *credential* from the environment (option → `NEON_API_KEY` → credentials file) but resolves **no host**, so `createRealNeonApi` falls back to `DEFAULT_NEON_API_BASE_URL = https://console.neon.tech/api/v2`. `createRealNeonApi` already accepts `baseUrl` — nothing fills it.

This PR gives the host the same treatment as the credential, at the same single choke point. A follow-up neonctl PR threads `props.apiHost` (the resolved `--api-host` flag) into these options; the env-var path works for all consumers with this PR alone.

### `@neondatabase/config`
- `createNeonApiFromOptions(operation, { apiKey?, apiHost? })` now resolves the API host: explicit `apiHost` option → `NEON_API_HOST` env → production default. Resolution happens at call time (not module load). Each source is normalized independently, so a blank/whitespace option falls through to the env var — symmetric with the existing `apiKey` chain.
- New module-private `normalizeApiHost`: trims surrounding whitespace and trailing slashes, treats empty as unset (guards `NEON_API_HOST=https://.../api/v2/` from producing a subtly different base URL).
- `createRealNeonApi` is untouched — it stays the explicit low-level constructor taking `baseUrl`.
- `NEON_API_HOST` added to the test clean-env list (`test-utils.ts`).
- Tests (`auth.test.ts`): option wins over env; env fallback; neither set → prod default (regression guard); trailing-slash/whitespace normalization; blank option falls through to env.

### `@neondatabase/config-runtime`
- `PullConfigOptions`, `PushConfigOptions`, and `ConfigOperationOptions` gain optional `apiHost` (doc: "Neon API base URL. Falls back to `NEON_API_HOST`, then production."), forwarded to `createNeonApiFromOptions` by `pullConfig`, `pushConfig`, and the `inspect`/`plan`/`apply` wrappers.
- New shared test `apihost-forwarding.test.ts`: sentinel-mock pattern asserting each entry forwards `{ apiKey, apiHost }` to the right callee (`inspect`→`pullConfig`, `plan`/`apply`→`pushConfig`).
- `pushConfig` header TSDoc corrected: it no longer claims to read no `NEON_*` env vars (the API credential/host resolution documented on `apiKey`/`apiHost` does).
- `createBranch`/`CreateBranchOptions` deliberately untouched: neonctl's callers use their own host-correct client; programmatic users are covered by the `NEON_API_HOST` env fallback.

### `@neondatabase/env`
- `FetchEnvOptions` gains optional `apiHost`, forwarded by `fetchEnv`. The field is named `apiHost` (not `baseUrl`) because this package already uses `baseUrl` for the Neon Auth publishable URL — the doc comment disambiguates ("Neon **management** API base URL (not the Auth base URL)").
- Its private `createApiFromOptions` normalized to the same spread style as config-runtime (behavior-identical for existing callers).
- New test `env.apihost.test.ts` (mocks `@neondatabase/config/v1`, the actual import specifier).

### DIFF
--- a/.changeset/honor-neon-api-host.md
+++ b/.changeset/honor-neon-api-host.md
@@ -1,0 +1,7 @@
+---
+"@neondatabase/config": patch
+"@neondatabase/config-runtime": patch
+"@neondatabase/env": patch
+---
+
+Honor `NEON_API_HOST` / the new `apiHost` option when building the default Neon API client. `createNeonApiFromOptions` now resolves the host (explicit `apiHost` option → `NEON_API_HOST` env → production default), and `pullConfig`, `pushConfig`, `inspect`/`plan`/`apply`, and `fetchEnv` accept and forward an optional `apiHost`.
--- a/packages/config-runtime/package.json
+++ b/packages/config-runtime/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@neondatabase/config-runtime",
-	"version": "0.4.1",
+	"version": "0.4.2",
 	"description": "Imperative runtime for @neondatabase/config: inspect/plan/apply a neon.ts policy against the Neon API and bundle + deploy Neon Functions. Pulls in esbuild; import this from CLIs and CI, not from neon.ts.",
 	"keywords": [
 		"neon",
--- a/packages/config-runtime/src/lib/apihost-forwarding.test.ts
+++ b/packages/config-runtime/src/lib/apihost-forwarding.test.ts
@@ -26,12 +26,12 @@ describe("pullConfig — apiHost forwarding", () => {
 				projectId: "p",
 				branchId: "br-1",
 				apiKey: "napi_k",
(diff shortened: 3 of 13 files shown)


andrelandgraf (approved):
Thank you!

PR #186, opened 2026-06-11
fix(env): emit NEON_AI_GATEWAY_BASE_URL as the bare branch gateway host

guillaumervls:
## What

`toEntries` (used by `neon-env run` / `export`) now emits `NEON_AI_GATEWAY_BASE_URL` as the **bare branch gateway host** (`scheme://host`, no path, via `new URL(baseUrl).origin`) instead of the OpenAI base URL with `/ai-gateway/openai/v1` rewritten to `/ai-gateway`.

## Why

The `@ai-sdk/neon` provider (https://github.com/vercel/ai/pull/15997) appends the full `/ai-gateway/<dialect>/…` routes to this value itself, so a value already ending in `/ai-gateway` would produce `…/ai-gateway/ai-gateway/…` → 404. The host-only value is what the provider's contract expects.

`OPENAI_BASE_URL` and `NEON_AI_GATEWAY_TOKEN` are unchanged.

## Changes

- `toEntries`: `NEON_AI_GATEWAY_BASE_URL` = `new URL(ai.baseUrl).origin`; removed the now-unused `AI_GATEWAY_BASE_PATH` const; updated the `NEON_ENV_VAR_KEYS.aiGateway` doc comment.
- Test pins the exact bare-host literal (`https://x.neon.build`).
- Updated the pending `ai-gateway-branch-host` changeset to describe the new value, and fixed a stale line in the pending `branch-credentials` changeset that still described `OPENAI_BASE_URL` as derived from the API origin (it's the branch gateway host).


andrelandgraf (approved):
Thanks!

PR #194, opened 2026-06-12
fix(env): use full credential token_id for AWS_ACCESS_KEY_ID (object storage broken)

andrelandgraf:
## Summary

`neon env pull` / `neon dev` write an `AWS_ACCESS_KEY_ID` that the S3-compatible storage endpoint **rejects**, so object storage is unusable from a pulled `.env` even though the bucket exists and the file looks complete.

`fetchEnv`'s `resolveCredentialSecrets` set `accessKeyId` to the credential's display-only short id (`token_id_short`, 12 chars) instead of its full `token_id` (`nak_live_…`, 41 chars). The storage endpoint authenticates with the **full** token id, so every request fails:

```
InvalidAccessKeyId: The AWS Access Key Id you provided does not exist in our records.
```

Fix: emit `minted.tokenId`. `AWS_SECRET_ACCESS_KEY` and the AI Gateway / Postgres / Auth vars are unchanged.

## Reproduced + verified on staging (us-east-2)
Full flow `bootstrap → link → config apply → env pull` against a real project with `preview.buckets`:

- **Before:** `AWS_ACCESS_KEY_ID` = `79feadb0c832` (len 12) → S3 `PutObject`/`ListObjectsV2` fail with `InvalidAccessKeyId`.
- **After:** `AWS_ACCESS_KEY_ID` = `nak_live_…` (len 41) → S3 `PutObject` + `ListObjectsV2` **succeed**.

(Confirmed independently by retrying S3 with the full `token_id` from the credentials API — it works; the short id does not.)

## Release
- Bumps `@neondatabase/env` `0.5.0 → 0.5.1` and adds a changeset (patch), so a release can be cut on merge.

## Test plan
- [x] `@neondatabase/env` unit suite (57 tests) + `tsc --noEmit` clean
- [x] e2e on staging: re-pulled env after the fix → S3 PUT/LIST succeed

### DIFF
--- a/.changeset/env-storage-access-key-id.md
+++ b/.changeset/env-storage-access-key-id.md
@@ -1,0 +1,5 @@
+---
+"@neondatabase/env": patch
+---
+
+Fix object storage being unusable from a pulled `.env`. `fetchEnv` (and therefore `neon env pull` / `neon dev`) set `AWS_ACCESS_KEY_ID` to the credential's display-only short id (`token_id_short`) instead of its full `token_id` (e.g. `nak_live_…`). The S3-compatible storage endpoint authenticates with the **full** token id, so every request failed with `InvalidAccessKeyId` ("The AWS Access Key Id you provided does not exist in our records") even though the `.env` looked complete and the bucket existed. Now emits the full `token_id`, so the AWS SDKs work from the pulled env out of the box. `AWS_SECRET_ACCESS_KEY` and the AI Gateway / Postgres vars are unchanged.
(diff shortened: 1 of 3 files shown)


andrelandgraf:
Closing as a duplicate — the same fix already landed on `main` via #197 (`fix(env): use full token id for S3 AWS_ACCESS_KEY_ID`), with its own changeset. Independently reproduced + verified this on staging (S3 PUT/LIST fail with the short `token_id_short`, succeed with the full `token_id`), so #197 resolves it. No action needed here.

PR #268, opened 2026-07-02
feat(sdk): regenerate client with branch recovery operations

aluuu:
## PR Checklist

- [X] Addresses an existing open issue: pending release of branch recovery API

## Overview

Regenerate @neon/sdk from the latest published Neon OpenAPI spec, adding recoverProjectBranch, include_deleted on listProjectBranches, hard_delete on deleteProjectBranch, and Branch.recovery. Update the ergonomic-layer drift guard snapshot with the 19 new raw operations.

Co-authored-by: Isaac


ivanoglo (changes requested):
Did /review from Isaac - 0 findings. Checked that pnpm spec:pull and I got 9 ilnes difference from this Pr's neon-openapi.json.
1. SnapshotUpdateRequest gained an expires_at field.
2. /users/me/organizations description has expanded to "When called with an organization- or project-scoped API key (which is not\ntied to a user), this returns the single organization that owns the key".
After corepack pnpm generate - files sdk.gen.ts and types.gen.ts are changed with corresponding lines.

I guess these are minor and not blocking, since the field `expires_at?: string | null;` is optional. But these changes are better to have, because once sdk is released - people will start to use it and they won't have that expires_at parameter, which they might have needed.

vitest run src/neon/coverage.test.ts passes.

It's a safe PR, because we just add more functionality to sdk and don't have to think about backward compatibility.

andrelandgraf:
Should already be in latest (and sorry I didn't see this PR in time!)

PR #269, opened 2026-07-02
feat(cli): add ability to list deleted branches and recover deleted branch

aluuu:
## why?

The Neon public API now supports **branch recovery**: a deleted branch can be recovered within a 7-day window instead of being lost. This change adds `neonctl` support for that feature so users can recover a deleted branch, list deleted branches, see the recovery deadline and permanently delete a branch.

## what?

- `neonctl branches recover <id|name>` — new subcommand
- `branches list --include-deleted` — includes recoverable deleted branches; sends `include_deleted=false` explicitly when the flag is absent.
- Deleted-branch rendering — in `list` table output a deleted branch shows `Current State = deleted` plus a `Recoverable Until` column (`recoverable_until`).
- `branches delete --hard-delete` — permanently deletes immediately, skipping the 7-day recovery window (`hard_delete=true`).


packages/cli/src/commands/branches.test.ts
```diff
@@ -275,6 +293,49 @@ describe("branches", () => {
 		]);
 	});
 
+	test("delete by id with hard-delete", async ({ testCliCommand }) => {
```
ivanoglo, line 296: Why are we only testing delete by id and recover by id and name? We can delete by name too, we should test that too.

ivanoglo (approved):
I have one minor comment. I tried to test it locally, but seems like it does not work. Maybe because this feature is not enabled. How can I test this functionality (I don't have access to posthog)?

andrelandgraf (approved):

PR #283, opened 2026-07-08
feat(cli): add `neon inspect db` diagnostic commands

sav-maya:
## Description

Adds a `neon inspect db <subcommand>` command group to the CLI. A set of
read-only Postgres diagnostics for quickly checking a branch's database health.

### What's added

Fourteen diagnostic subcommands, each running one read-only query against
Postgres' built-in statistics/catalog views:

| Subcommand | What it reports | Source views | Needs extension |
| --- | --- | --- | --- |
| `table-sizes` | Size of each table (incl. TOAST), largest first | `pg_table_size` | — |
| `index-sizes` | Size of each index, largest first | `pg_relation_size` | — |
| `unused-indexes` | Non-unique, non-PK indexes with few scans | `pg_stat_user_indexes` | — |
| `seq-scans` | Sequential-scan count per table | `pg_stat_user_tables` | — |
| `long-running-queries` | Queries running longer than 5 minutes | `pg_stat_activity` | — |
| `locks` | Locks held, with the acquiring query and its age | `pg_locks` + `pg_stat_activity` | — |
| `vacuum-stats` | Last (auto)vacuum, dead tuples, autovacuum expectation | `pg_stat_user_tables` + `pg_class` | — |
| `bloat` | Statistical table-bloat estimate (ratio + wasted bytes) | `pg_stats` + `pg_class` | — |
| `replication-slots` | Slot kind, status, client, restart/confirmed-flush LSNs, lag | `pg_replication_slots` + `pg_stat_replication` | — |
| `subscriptions` | Per-table logical replication progress on a subscriber | `pg_subscription_rel` | — |
| `outliers` | Queries by cumulative execution time | `pg_stat_statements` | `pg_stat_statements` |
| `calls` | Most frequently called queries | `pg_stat_statements` | `pg_stat_statements` |
| `lfc-hit-rate` | Local File Cache hit rate (the effective cache on Neon) | `neon_get_lfc_stats()` | `neon` |
| `working-set` | Estimated working set vs LFC size across 1m/5m/15m/1h windows | `approximate_working_set_size_seconds()` | `neon` |

### How it works

- **Command wiring** (`commands/inspect.ts`): follows the existing nested-command
  pattern (`inspect` → `db` → one subcommand per query), registered in
  `commands/index.ts`. Results are printed through the existing
  `writer(props).end(rows, { fields })` helper, so `--output table|json|yaml`
  works unchanged, with a friendly empty-set message where a blank table would
  otherwise print.
- **Query definitions** (`utils/inspect_queries.ts`): each subcommand's SQL,
  output columns (display order), empty-set message, and optional
  `requiresExtension` live in one keyed table.
- **DB access + connection resolution** (`utils/inspect_db.ts`): a self-contained
  `resolveConnectionUri(props)` (endpoint → role → password → database → URL,
  honoring `--project-id`, `--branch`, `--role-name`, `--database-name`,
  point-in-time, pooling, and SSL mode) that mirrors how `connection-string`
  resolves a branch — so `inspect db` connects the same way, honoring the linked
  `.neon` context (`neon link` / `set-context`) via the CLI's global context
  middleware, without touching any existing file. The same module then runs the
  query over the CLI's **embedded Postgres wire client** (`PgConnection` +
  `parseConnectionUri`) — no new `pg` dependency — and maps each result row to an
  object keyed by column name for the writer. Connection failures surface a
  Postgres-specific error instead of being mislabeled as a Neon API network
  error. For gated queries it first checks `pg_extension` and fails with an
  actionable `CREATE EXTENSION …` hint if the extension is missing.
- **`--db-url`**: inspect any Postgres directly via a connection string,
  bypassing the Neon API entirely (project/branch resolution is skipped).

### Connecting

There are three ways to point `inspect db` at a database — it resolves
connections the same way the rest of the CLI does.

**1. Any Postgres via `--db-url`** — the inspection queries are Postgres-agnostic,
so you can run them against any database (not just Neon) by passing a connection
string. This bypasses the Neon API entirely:

```bash
neon inspect db bloat --db-url postgresql://postgres:postgres@localhost:5432/postgres
```

**2. A linked Neon project (`neon link` / set-context)** — once the project is
linked (or a context is set), the CLI connects automatically whenever you're in
the project folder, and you no longer need `--db-url`:

```bash
neon link                 # or: neon set-context --project-id <project-id>
neon inspect db bloat
```

**3. Explicit flags** — override resolution per-invocation:

```bash
neon inspect db bloat \
  --project-id <project-id> --branch <branch> \
  --role-name <role> --database-name <db>
```

Output format is controlled by the shared `--output table|json|yaml` flag, e.g.
`neon inspect db bloat --output json`.

### Notes on specific commands

- `outliers` / `calls` require the `pg_stat_statements` extension, and
  `lfc-hit-rate` / `working-set` require the `neon` extension; all four are
  guarded at runtime (an actionable `CREATE EXTENSION …` hint if missing).
  Everything else needs no extension.
- `lfc-hit-rate` / `working-set` are Neon-specific. `lfc-hit-rate` reads the
  hit/miss counters from `neon_get_lfc_stats()` rather than the
  `neon_stat_file_cache` / `neon_lfc_stats` views, because those views are owned
  by `cloud_admin` and are not always readable by end-user roles, whereas the
  function is `EXECUTE`-able by the default role. `working-set` compares
  `approximate_working_set_size_seconds()` at 1m/5m/15m/1h against
  `neon.file_cache_size_limit`; `exceeds_lfc = yes` is the signal to size compute up.
- `bloat` is a statistical estimate (from `pg_stats`), not an exact measurement.
- `replication-slots` lag is recovery-aware (`pg_last_wal_receive_lsn()` on a
  standby, `pg_current_wal_lsn()` otherwise) and falls back to `restart_lsn` when
  a slot has no `confirmed_flush_lsn` (physical slots). The `status` column is a
  single derived value: the `pg_stat_replication` WAL-sender state
  (`streaming`/`catchup`/…) when a live walsender is attached, else
  `active (no walsender)` for non-walsender consumers (logical decoding / Neon's
  internal slot), else `inactive`. The `slot_kind` column classifies each slot:
  `physical` (streaming replicas / Neon's internal slot), `initial-sync
  (tablesync)` for the temporary per-table slots Postgres creates during a
  subscription's first copy (named `pg_<subid>_sync_<relid>`), and `CDC (apply)`
  for the subscription's steady-state logical slot.
- `subscriptions` is subscriber-side: it only returns rows on a database that is
  itself a logical-replication **subscriber** (has `CREATE SUBSCRIPTION` set up).
  On a non-subscriber database it correctly prints "No subscriptions found." Full
  output (per-table `initialize`/`copying`/`ready` states) still needs to be
  verified on an actual subscriber — e.g. a Neon branch subscribed to an upstream
  publisher during a logical-replication migration.

### Scope

- **Purely additive.** The branch only adds new files under `packages/cli/src/`
  (`commands/inspect.ts`, `commands/inspect.test.ts`, `utils/inspect_queries.ts`,
  `utils/inspect_db.ts`). The single unavoidable edit to
  an existing file is the 2-line command registration in `commands/index.ts` —
  no other existing file is modified (`connection-string`, the test harness, etc.
  are untouched).
- No structural or dependency changes: no `package.json`, workspace, `biome.json`,
  or `tsconfig` changes; no new runtime deps.

### Testing

- `pnpm --filter neonctl build` and `pnpm --filter neonctl lint` pass.
- `commands/inspect.test.ts` added (hermetic `--db-url` cases for all subcommands).
- All subcommands verified against a live linked branch (and against a plain
  Postgres via `--db-url`), in table and JSON output; the `pg_stat_statements`
  guard verified both with and without the extension installed;
  `replication-slots` verified against real logical + physical slots.

### DIFF
--- a/packages/cli/src/commands/inspect.ts
+++ b/packages/cli/src/commands/inspect.ts
@@ -1,0 +1,102 @@
+import type yargs from "yargs";
+import type { BranchScopeProps, CommonProps } from "../types.js";
+import { fillSingleProject } from "../utils/enrichers.js";
+import { resolveConnectionUri, runInspectQuery } from "../utils/inspect_db.js";
+import {
+	INSPECT_QUERIES,
+	type InspectQuery,
+	type InspectSubcommand,
+} from "../utils/inspect_queries.js";
+import { writer } from "../writer.js";
+
+export const command = "inspect";
+export const describe = "Inspect a database's health and configuration";
+export const aliases = ["inspection"];
+
+type InspectProps = BranchScopeProps & {
+	branch?: string;
+	roleName?: string;
+	databaseName?: string;
+	dbUrl?: string;
+};
+
+/**
+ * `fillSingleProject` hits the API to auto-resolve a project. When the user
+ * passes `--db-url` we bypass the Neon API entirely, so skip it.
+ */
+const fillSingleProjectUnlessDbUrl = async (
+	props: CommonProps & { projectId?: string; orgId?: string; dbUrl?: string },
+) => {
+	if (props.dbUrl) {
+		return props;
+	}
+	return fillSingleProject(props);
+};
+
+const runSubcommand = async (name: InspectSubcommand, props: InspectProps) => {
+	const query: InspectQuery = INSPECT_QUERIES[name];
+
+	const connectionUri =
+		props.dbUrl ?? (await resolveConnectionUri(props)).connectionUri;
+
+	const rows = await runInspectQuery(connectionUri, query.sql, {
+		requiresExtension: query.requiresExtension,
+	});
+	writer(props).end(rows, {
+		fields: query.fields as readonly (keyof (typeof rows)[number])[],
+		emptyMessage: query.emptyMessage,
+	});
+};
+
+const dbBuilder = (argv: yargs.Argv) => {
+	let builder = argv
+		.usage("$0 inspect db <sub-command> [options]")
+		.options({
+			"project-id": {
+				describe: "Project ID",
+				type: "string",
+			},
+			branch: {
+				describe: "Branch ID or name",
+				type: "string",
+			},
+			"database-name": {
+				describe: "Database name",
+				type: "string",
+			},
+			"role-name": {
+				describe: "Role name",
+				type: "string",
+			},
+			"db-url": {
+				describe:
+					"Inspect any Postgres via a connection string, bypassing the Neon API (project/branch resolution is skipped)",
+				type: "string",
+			},
+		})
+		.middleware(fillSingleProjectUnlessDbUrl as any);
+
+	for (const name of Object.keys(INSPECT_QUERIES) as InspectSubcommand[]) {
+		builder = builder.command(
+			name,
+			INSPECT_QUERIES[name].describe,
+			(yargs) => yargs,
+			(args) => runSubcommand(name, args as unknown as InspectProps),
+		);
+	}
+
+	return builder;
+};
+
+export const builder = (argv: yargs.Argv) =>
+	argv
+		.usage("$0 inspect <sub-command> [options]")
+		.command(
+			"db",
+			"Run a diagnostic query against a branch's Postgres",
+			dbBuilder,
+		);
+
+export const handler = (args: yargs.Argv) => {
+	return args;
+};
--- a/packages/cli/src/commands/__snapshots__/inspect.test.ts.snap
+++ b/packages/cli/src/commands/__snapshots__/inspect.test.ts.snap
@@ -1,0 +1,21 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`inspect db > bloat --db-url bypasses the API and reports a Postgres connection error 1`] = `""`;
+
+exports[`inspect db > cache-hit --db-url bypasses the API and reports a Postgres connection error 1`] = `""`;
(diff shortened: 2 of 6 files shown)


andrelandgraf (approved):

packages/cli/src/commands/inspect.ts
```diff
@@ -0,0 +1,102 @@
...
+			},
+			"database-name": {
+				describe: "Database name",
+				type: "string",
+			},
+			"role-name": {
+				describe: "Role name",
+				type: "string",
+			},
+			"db-url": {
```
vadim2404, line 71: I'm not sure that we need to inspect any arbitrary Postgres. Let's focus only on things we control

sav-maya: I can pull `--db-url` if we want to keep this Neon-only, but I'd lean toward keeping it.

I think we should be building great Postgres tooling, not just Neon tooling. If we make inspect genuinely useful against any Postgres, it becomes something people reach for regardless of where their database lives, and that's how they end up trusting us with the database too. Migrations are the obvious wedge: someone inspects their current Postgres, likes the tool, and Neon is right there when they're ready to move.

If scope is the worry, we make it a secondary flag and keep the linked project flow as the main path.

Fine to cut it if you feel strongly, but I think meeting people where their Postgres already is only helps us.

PR #322, opened 2026-07-20
fix(env): prefer default neondb when a branch has multiple databases

thisistonydang:
## Summary

`fetchEnv` previously threw `AmbiguousBranchAuth` as soon as a branch had more than one database — so `neonctl link` / `neonctl env pull` **failed** on a branch that had `neondb` alongside another database (e.g. `my-database` + `neondb`).

Database auto-pick is now:

- Prefer Neon's default **`neondb`** if present (covers the common "added a second database next to neondb" case).
- Else the **sole** database.
- Else (several databases, none named `neondb`) **fail loudly** — auto-picking one would be perceived as random and is bad DX.

The failure message no longer implies a `databaseName` CLI flag (there is none on `link` / `env pull` / `neon-env`); it suggests renaming a database to `neondb` or keeping a single database, and mentions `databaseName` only for callers using `fetchEnv` directly.

`fetchEnv` stays a pure, reusable library — no CLI/logging concerns are threaded into it.

## Changes

- `packages/env/src/lib/env.ts` — `pickDatabaseName` prefers `neondb`, else the sole DB, else throws with actionable guidance; updated `databaseName` JSDoc.
- `packages/env/src/lib/env.test.ts` — auto-pick tests (prefers `neondb`, sole DB, throws when several + no `neondb`, explicit `databaseName` wins).
- `packages/env/README.md` — document role/database selection.
- `.changeset/` — `@neon/env` minor + `neonctl` patch.

## Review notes

Addresses review feedback on the first revision:
- Removed the `onNotice` callback (kept `fetchEnv` free of CLI concerns).
- Removed the "just pick one" fallback when there's no `neondb` — fail loudly instead.

## Test plan

- [x] `@neon/env` unit tests (103 passing).
- [x] `neonctl` env-path tests (`dev/env.test.ts`, `commands/env.test.ts` — 21 passing).
- [x] Verified end-to-end: built CLI + `neonctl link` on a real branch with `my-database` + `neondb` pulls env for `neondb`.
- [ ] CI: full `pnpm lint:ci` / `pnpm test:ci` (couldn't run repo-wide locally — corepack offline).

### DIFF
--- a/packages/env/src/lib/cli/commands.ts
+++ b/packages/env/src/lib/cli/commands.ts
@@ -198,8 +198,6 @@ async function loadConfigAndFetchEnv(
 		env: { ...process.env, ...fileEnv },
 		...(ctx.api ? { api: ctx.api } : {}),
 		...(options.apiKey ? { apiKey: options.apiKey } : {}),
-		// Info notices (e.g. auto-picked database) go to stderr so `export`'s stdout stays clean.
-		onNotice: (message) => process.stderr.write(`${message}\n`),
 	});
 }
 
--- a/packages/env/src/lib/cli/commands.ts
+++ b/packages/env/src/lib/cli/commands.ts
@@ -198,6 +198,8 @@ async function loadConfigAndFetchEnv(
 		env: { ...process.env, ...fileEnv },
 		...(ctx.api ? { api: ctx.api } : {}),
 		...(options.apiKey ? { apiKey: options.apiKey } : {}),
+		// Info notices (e.g. auto-picked database) go to stderr so `export`'s stdout stays clean.
+		onNotice: (message) => process.stderr.write(`${message}\n`),
 	});
 }
 
--- a/packages/env/src/lib/env.ts
+++ b/packages/env/src/lib/env.ts
@@ -32,6 +32,13 @@ import { z } from "zod";
 const NEON_DEFAULT_OWNER_ROLE = "neondb_owner";
 
 /**
+ * Neon's default database, created with every project. When a branch has several databases
+ * and none was requested, this is preferred for the `DATABASE_URL` so the common case (a
+ * user added a second database next to `neondb`) auto-picks without asking.
+ */
+const NEON_DEFAULT_DATABASE = "neondb";
+
+/**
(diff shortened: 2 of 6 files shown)


packages/env/src/lib/cli/commands.ts
```diff
@@ -198,6 +198,8 @@ async function loadConfigAndFetchEnv(
 		env: { ...process.env, ...fileEnv },
 		...(ctx.api ? { api: ctx.api } : {}),
 		...(options.apiKey ? { apiKey: options.apiKey } : {}),
+		// Info notices (e.g. auto-picked database) go to stderr so `export`'s stdout stays clean.
+		onNotice: (message) => process.stderr.write(`${message}\n`),
```
andrelandgraf: Remove this - `fetchEnv` is not only used in the CLI but exposed so you can use it in your own dev scripts and potentially even runtime code. `onNotice` introduces CLI concerns into the reusable `fetchEnv` func.

packages/env/src/lib/env.ts
```diff
@@ -908,25 +925,36 @@ function pickDatabaseName(
...
+			onNotice?.(
+				`Branch ${branch.name} has ${databases.length} databases (${names.join(", ")}); using the default '${NEON_DEFAULT_DATABASE}'.`,
+			);
+		}
+		return neondb.name;
+	}
+
+	if (databases.length === 1) return databases[0].name;
+
+	// No `neondb` and several databases: pick deterministically rather than fail. Prefer one
```
andrelandgraf: If there is no `neondb` and there are several databases; fail loudly. 

This is ambiguous to the user - will be perceived as random. If we give them a "random" db that's not good DX. Better to fail loudly and let them know we expected `neondb` or just one db.

andrelandgraf (commented):
Needs work:

- remove `onNotice`
- remove "just pick first!" logic if `neondb` isn't present

thisistonydang:
@andrelandgraf PR has been updated per your comments: `neondb` is always auto-picked if present, else a sole database, else throw if there are multiple dbs. The thrown error has also been updated to no longer imply a `databaseName` CLI flag that doesn't exist :)

andrelandgraf (approved):

PR #331, opened 2026-07-25
feat(cli): make neon the CLI package and neonctl a compatibility shim

andrelandgraf:
## Problem

The Neon CLI is published twice. The release workflow packs `packages/cli` as `neonctl`, then re-packs the byte-identical build under the alias name `neon`. Two independent tarballs, one codebase, no relationship between them on npm.

That has one consequence we care about: installing `neonctl` never resolves `neon`. Every install of the old name is a download the new name doesn't see, so the package we're asking people to converge on undercounts by exactly the traffic its own command is serving. It also means two sets of dependency metadata and two provenance trails for one binary.

## Approach

`packages/cli` is now published as **`neon`** and is the only package that carries CLI source, dependencies, and standalone binaries. **`packages/neonctl`** is a new package containing a `package.json`, a `README.md`, and this executable:

```js
#!/usr/bin/env node

await import("neon/cli");
```

Its only dependency is `neon` (`workspace:*`, rewritten to the exact released version by `pnpm pack`). `neon` gained an `exports` map so `neon/cli` is a supported public entry point rather than a deep path into `dist/`. A Changesets `fixed` group ties the two packages together so they can never be released at different versions.

The shim declares **both** command names against that one file:

```json
"bin": {
  "neonctl": "bin/cli.js",
  "neon": "bin/cli.js"
}
```

`process.argv[1]` is the link the user invoked rather than the file it resolves to, so each command brands its own help without the shim passing anything along. Keeping both names is what makes this a drop-in replacement for the package it supersedes — and it is what keeps Homebrew working, see below.

Three properties fall out of this:

1. **Installing `neonctl` downloads `neon`.** npm counts tarball responses, so any uncached `npm i neonctl`, `npx neonctl`, or `brew install neonctl` now registers a download against `neon` too. (Nothing can count *executions* of an already-installed CLI — that isn't something npm measures.)
2. **One codebase.** The shim has no copied CLI code and no build step. It cannot drift.
3. **`neonctl` behaves exactly as before.** An in-process `import` — not a spawned child — preserves argv, stdio, signals, cwd, and exit codes, and both commands stay available.

## What each distribution channel does after this

| Channel | Result |
| --- | --- |
| `npm i -g neon` | The recommended install. Provides `neon`. |
| `npm i -g neonctl` | Provides `neonctl` and `neon`, as today, and additionally downloads `neon`. |
| `brew install neonctl` | **Unchanged formula.** Provides both commands. |
| GitHub releases | `Neon CLI <version>`, tagged `neon@<version>`, with `neon-<platform>` assets. |

### Homebrew deserves a callout

`brew install neonctl` is [a homebrew-core formula](https://github.com/Homebrew/homebrew-core/blob/main/Formula/n/neonctl.rb) that builds from the npm `neonctl` tarball, links whatever executables it declares, generates completions for `neonctl` *and* `neon`, and prunes the bundled esbuild out of `lib/node_modules/neonctl/node_modules`. Homebrew's bot bumps it within a day of each publish.

A shim with only `neonctl` would have failed that formula's completion step and dropped the `neon` command for brew users, forcing a homebrew-core PR timed between the two npm publishes. Declaring both `bin` entries avoids all of it — no formula change, no coordinated release step.

Worth stating plainly, because it inverts how the package should be thought of: homebrew-core's `neon` is already the [neon HTTP/WebDAV library](https://formulae.brew.sh/formula/neon), so `neonctl` is the name Homebrew has for the Neon CLI no matter what npm calls it. The npm `neonctl` package is load-bearing, not vestigial — hence a supported compatibility package rather than a deprecated alias. `AGENTS.md` and the release skill now warn against narrowing its `bin` map.

## Verification

Beyond CI, I exercised the real artifacts rather than reasoning about them.

**Packed tarballs, installed like a user would:** `neonctl-<version>.tgz` is 4 files declaring `{ neon: '<same version>' }` (the workspace protocol resolves at pack time). `neonctl --version` and `neon --version` agree, `neonctl --help` prints `neonctl <command> [options]`, `neon --help` prints `neon <command> [options]`, and npm dedupes to a single `neon` with no nested copy.

**Homebrew's install steps, replayed against the packed shim:**

```
$ npm install --global --prefix ./prefix ./neonctl-<version>.tgz   # system "npm", "install", *std_npm_args
$ ls prefix/bin                                                    # bin.install_symlink libexec.glob("bin/*")
neon
neonctl
$ prefix/bin/neonctl completion bash | head -1                     # generate_completions_from_executable
#compdef neonctl
$ prefix/bin/neon completion bash | head -1
#compdef neon
$ ls prefix/lib/node_modules/neonctl/node_modules | grep esbuild   # the formula's prune path
@esbuild
esbuild
```

The `neon` dependency's transitive dependencies hoist to exactly the level the formula prunes, so the bundled esbuild is still removed in favor of the `esbuild` formula.

**Standalone binaries:** `pnpm --filter neon bundle` produces `neon-linux-arm64`, `neon-linux-x64`, `neon-macos-x64`, `neon-win-x64.exe`, matching the `bundle/neon-*` glob in the companion workflow PR. `pkg.js` needed one fix, included here: it deletes the new `exports` map for the standalone build, since the bundle has no `dist/` for those subpaths to resolve against.

**Versioning:** `pnpm exec changeset version` bumps both packages together, confirming the fixed group.

The shim's tests (`packages/neonctl/test/shim.test.ts`) run both commands through links named the way npm and Homebrew create them, and assert version and branding for each. They use Vitest like the rest of the repo, because root CI appends `--coverage` to every package's `test:ci`; pnpm's topological ordering builds `neon` before they run.

## Release impact

Requires the companion workflow change: [databricks/secure-public-registry-releases-eng#187](https://github.com/databricks/secure-public-registry-releases-eng/pull/187), which stops repacking the CLI under a second name and moves the binaries plus the GitHub release to the `neon` dispatch. **Merge that PR first** — until it lands, a release from this branch would publish the shim under the `neon` name.

Publish order is `@neon/config` → `@neon/config-runtime` / `@neon/env` → `neon` → `neonctl`, one dispatch each. The `neon`-before-`neonctl` half matters on every release: the shim's dependency is an exact pin, so flipping the order leaves a window where `npm i neonctl` can't resolve and a Homebrew bump landing in that window can't build. Documented in `AGENTS.md`, the release skill, and `docs/neonctl-compatibility-shim.md`.

Nothing else needs a release-time step. `neon-init` is unaffected: it probes a globally installed `neonctl --version` against `npm view neonctl version`, which stays consistent because the fixed group keeps both packages on one version.

## Reviewing this

`docs/neonctl-compatibility-shim.md` is the argument in full, including the five alternatives rejected: publishing the same tarball twice (the status quo), spawning the `neon` executable, pointing `bin` at a path inside `node_modules`, installing `neon` at runtime, and npm dependency aliases.

The only behavior change for existing users: the `neon` package alone provides just the `neon` command. Someone who ran `npm i -g neon` and types `neonctl` needs the `neonctl` package (which is also what Homebrew installs) or the `neon` command. `neonctl` itself is unaffected and is not deprecated.


andrelandgraf:
## Verified the other two release channels

### GitHub releases — works, no further change

Ran the cross-compile locally (`pnpm --filter neon bundle`). Because `pkg` names artifacts after the package, they now come out as:

```
neon-linux-arm64  neon-linux-x64  neon-macos-x64  neon-win-x64.exe
```

That matches the `bundle/neon-*` glob and `cp` in the companion workflow PR, and the `.exe` is covered by the same glob. The release is tagged `neon@<version>` on this repo. `pkg.js` needed one fix, already in this PR: it deletes the new `exports` map for the standalone build, since the bundle has no `dist/` for those subpaths to point at.

Binary download links in `packages/cli/README.md` now point here rather than the archived `neondatabase/neonctl` releases.

### Homebrew — needs a one-time formula PR

This one does **not** work unchanged. [`homebrew-core`'s `neonctl` formula](https://github.com/Homebrew/homebrew-core/blob/main/Formula/n/neonctl.rb) builds from the npm `neonctl` tarball and does `bin.install_symlink libexec.glob("bin/*")`, so brew users get both `neonctl` and `neon` today only because the fat package declared both executables.

I simulated the formula's install step against the packed compatibility tarball:

```
$ npm install --global --prefix ./prefix ./neonctl-2.36.0.tgz
$ ls prefix/bin
neonctl
$ ls prefix/lib/node_modules/neonctl/node_modules/.bin
... neon ...
```

The `neon` executable lands in `node_modules/.bin`, which Homebrew never exposes, so `bin/neon` doesn't exist and the formula's `generate_completions_from_executable(bin/"neon", …)` fails the build. Separately, its cleanup paths (`libexec/"lib/node_modules/neonctl/node_modules"`) would glob a directory that no longer describes the tree and silently stop pruning the bundled esbuild the formula declares a dependency on.

The fix is to build the formula from `neon` and symlink the legacy name. I verified that install path too — `neonctl --version` → `2.36.0`, `neonctl --help` → `neonctl <command> [options]`, `neon --help` → `neon <command> [options]`, and `neonctl completion bash` still emits `neonctl`-named completions. The formula keeps the name `neonctl`, because homebrew-core's `neon` is already the neon HTTP/WebDAV C library.

The exact diff and the timing constraint are now in `docs/neonctl-compatibility-shim.md`, with pointers from `AGENTS.md` and the release skill. Timing matters: the PR needs a published `neon@<version>` for its `url`/`sha256`, and Homebrew's bot has been bumping this formula within a day of each npm publish (2.35.1, 2.35.2, 2.36.0 on consecutive days), so it should land between the `neon` and `neonctl` dispatches. If the bot gets there first it opens a bump PR against the compatibility tarball that can't build — noisy, but not user-visible, since brew users stay on the last good version until it's fixed.

Also fixed here: the shim was authored at `2.37.0` with a pre-generated changelog. It now starts level with `packages/cli` at `2.36.0` and lets the fixed group bump both at release.

andrelandgraf:
## Correction: Homebrew needs no changes after all

My previous comment concluded that this required a timed homebrew-core PR. That was self-inflicted, and it's now fixed in the branch.

The formula breaks only because the shim declared a single `bin`, while the package it replaces declared two — and the shim's `bin` map is ours to choose. It now declares both names against one file:

```json
"bin": {
  "neonctl": "bin/cli.js",
  "neon": "bin/cli.js"
}
```

`process.argv[1]` is the link the user invoked rather than the file it resolves to, so each command still brands its own help without the shim passing anything along.

I replayed the formula's install steps against the packed shim tarball to confirm:

```
$ npm install --global --prefix ./prefix ./neonctl-2.36.1.tgz   # system "npm", "install", *std_npm_args
$ ls prefix/bin                                                 # bin.install_symlink libexec.glob("bin/*")
neon
neonctl
$ prefix/bin/neonctl --help | head -1
neonctl <command> [options]
$ prefix/bin/neon --help | head -1
neon <command> [options]
$ prefix/bin/neonctl completion bash | head -1                   # generate_completions_from_executable
#compdef neonctl
$ prefix/bin/neon completion bash | head -1
#compdef neon
$ ls prefix/lib/node_modules/neonctl/node_modules | grep esbuild # the formula's prune path
@esbuild
esbuild
```

Every step the formula performs still holds, including the `lib/node_modules/neonctl/node_modules` prune path — the `neon` dependency's transitive dependencies hoist to exactly that level, so the bundled esbuild is still removed in favor of the `esbuild` formula. So: no homebrew-core PR, no release-time coordination, and no lost `neon` command for brew users.

This also removes the only user-visible regression the change had. `npm i -g neonctl` now behaves exactly as it does today, providing both commands, and additionally pulls `neon`.

Worth stating plainly, since it inverts how the package should be thought of: **homebrew-core's `neon` is already the [neon HTTP/WebDAV library](https://formulae.brew.sh/formula/neon)**, so `neonctl` is the name Homebrew has for the Neon CLI no matter what npm calls it. The npm `neonctl` package is load-bearing infrastructure, not a vestigial alias — a good reason it is a supported compatibility package rather than a deprecated one. `AGENTS.md` and the release skill now warn against narrowing its `bin` map.

`neon-init` is unaffected for the same reason: it probes a globally installed `neonctl --version` and compares it to `npm view neonctl version`, which stays correct because the fixed group keeps the two packages on one version.

The companion workflow PR also now titles the release **`Neon CLI <version>`** while keeping the `neon@<version>` tag, so the releases page reads as the CLI rather than a package coordinate. Assets remain `neon-<platform>`.

PR #352, opened 2026-07-28
fix(cli, env): report credential secrets reused from .env, not re-fetched

philip:
## Problem
`neon env pull` listed storage / AI-gateway secrets (`AWS_ACCESS_KEY_ID`,
`AWS_SECRET_ACCESS_KEY`, `NEON_AI_GATEWAY_TOKEN`) under the same "Pulled N
variables" line as freshly-fetched vars. But those one-time secrets, when
already present, are reused from the existing `.env` — never re-fetched — so
the output looked like they'd been refreshed.

This bit most when copying a `.env.example` and running `pull`: the placeholder
secrets were reported as "Pulled" yet silently kept their example values,
leaving no signal that they weren't updated from Neon.

## Change
`fetchEnv` now reports how the branch credential was resolved via a new
`onCredential({ source, keys })` option. `env pull` uses it to split the output:

```
Pulled 5 Neon variables into .env: NEON_BRANCH, DATABASE_URL, ...
Reused 2 existing values as-is (not fetched/verified from Neon): AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY
```

Reporting only — no change to what's minted, reused, or written.

## Test
Added coverage for the `onCredential` signal (`@neon/env`) and the pull output
split (CLI). Verified against a real gateway+storage branch.

### DIFF
--- a/packages/cli/src/commands/env.test.ts
+++ b/packages/cli/src/commands/env.test.ts
@@ -25,7 +25,7 @@ import type {
 	NeonProjectSnapshot,
 	NeonRoleSnapshot,
 } from "@neon/config";
-import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { autoPullEnvAfterPin, type EnvPullProps, pull } from "./env.js";
 
@@ -200,6 +200,16 @@ class FakeNeonApi implements NeonApi {
 	}
 }
 
+/**
+ * A branch with object storage enabled (one bucket), so env-pull's tier-2 path (`pullConfig`
+ * -> `fetchEnv`) resolves the branch credential and emits the `AWS_*` storage vars.
+ */
+class StorageNeonApi extends FakeNeonApi {
+	override async listBranchBuckets(): Promise<NeonBucketSnapshot[]> {
+		return [{ name: "assets", accessLevel: "private" }];
+	}
+}
+
 /** Stand-in for the neonctl Api client; only branch resolution is exercised. */
 const fakeApiClient = {
 	listProjectBranches: async () => ({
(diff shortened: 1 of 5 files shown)


andrelandgraf:
Thanks for chasing this down, Philip — the diagnosis is spot on and the `.env.example` scenario in your description is exactly the right thing to have noticed. I dug into it and ended up going one layer deeper, so I've opened #353 and I'm closing this one in favour of it. Wanted to explain why rather than just swapping it out.

Your PR makes the output honest about the placeholders. I kept wanting the placeholders to not be there. The flow that `packages/ai-sdk-provider/.env.example` documents in its own first line — "Copy to .env and fill from `neonctl env pull`" — can't work today, and after this change it still can't; it just says so more clearly.

What convinced me it was fixable: I'd assumed the blocker was credential spam, and that avoiding it meant remembering the previous run's credential id somewhere, i.e. new state in `.neon`. That turns out to be wrong. `AWS_ACCESS_KEY_ID` **is** the credential's `tokenId`, and the AI Gateway token is minted as `nt_live_<tokenIdShort>_<secret>` — so the `.env` we're about to rewrite is already the bookkeeping. `listCredentials` and `revokeCredential` are already on the adapter, and `credentialScopesSatisfied` was written for precisely this decision (its docstring says so) and had no caller. So `env pull` can check the persisted secrets against the branch's live credentials, reuse them when they're real, and mint-and-revoke when they aren't — no spam, no new state.

Two things from your PR that I kept, credited:

- **`onCredential`.** Good idea and a good name. Mine reports `{ action, keys, revoked }` and is the same "let the caller name the affected keys without re-deriving which ones are credential-backed" argument you make in your comment. I'd initially written down that firing it before the storage fetch was a wart; once it can report a revocation I decided the opposite and fire it early on purpose, since a remote mutation has to reach the caller even if the resolve then throws. Your instinct was right there.
- **Splitting the reporting.** #353 doesn't need the split, but only because verified reuse makes the "Pulled N variables" line true for every key it lists — which is the outcome you were after.

A couple of loose ends worth having on the record, both of which #353 covers: `renderAgentPullNote` / `PullOutcome.written` still counted the reused secrets, so `link --agent` kept emitting the misleading number in JSON even after the human-facing fix; and this needed a changeset (`@neon/env` gains public API, and CLI output changes).

Also — while writing tests I found that `FakeNeonApi` was minting a colliding `tokenIdShort` for every credential (`` `c${seq}`.slice(0, 12) `` cuts off the digit that varies). Harmless until something looks a credential up by id, which is why no test ever caught it. Fixed in #353.

Genuinely useful report, and the repro in your description is what made the root cause easy to find. Happy to be talked out of any of the above if you think verify-on-pull is the wrong trade.

PR #353, opened 2026-07-28
fix(cli, env): verify branch credential secrets on env pull instead of trusting them

andrelandgraf:
Competing take on #352, which I closed in favour of this. Same bug, one layer down: that PR reports the placeholder problem, this one removes it — and takes the opportunity to make `fetchEnv` a function that only fetches.

## Problem

Object storage and AI Gateway secrets (`AWS_SECRET_ACCESS_KEY`, `NEON_AI_GATEWAY_TOKEN`) are returned by the API exactly once, at mint time. They aren't stored server-side and the list endpoint returns metadata only, so **the persisted copy is the only copy** — and "fetching" one means minting a new credential. `fetchEnv` therefore reused whatever its env source already carried, or every `pull`, `link`, `checkout` and `dev` start would leave another live credential on the branch.

That reuse was **presence-based**: any non-empty value counted as a real secret. So the flow `packages/ai-sdk-provider/.env.example` itself documents — "Copy to .env and fill from `neonctl env pull`" — could not work. I reproduced it against the fake API on `main`:

```
TOKEN AFTER PULL: nt_live_...
createCredential calls: 0
```

The placeholder was echoed back, counted in "Pulled N variables", and the resulting `.env` 401s on every gateway request. Presence-based reuse is equally blind to a credential revoked in the console and to one copied in from another branch.

## Why it was built that way, and what unblocks it

The reason not to just always mint is credential spam, and the reason not to fix *that* is that skipping the spam seems to require remembering the previous run's credential id — which would mean putting state in `.neon`.

It doesn't. **The persisted secrets already carry their own credential id.** `AWS_ACCESS_KEY_ID` *is* the credential's `tokenId` (the storage gateway authenticates against the full id), and the AI Gateway token is minted as `nt_live_<tokenIdShort>_<secret>`, where `tokenIdShort` is the identifier `listCredentials` reports. The `.env` we are about to rewrite is the bookkeeping. `listCredentials` and `revokeCredential` were already on the `NeonApi` adapter, and `credentialScopesSatisfied` was already written for this exact decision — its doc says so — and had no production caller.

So the fix isn't "always mint", it's **verify before reusing**: fake and outdated values both get replaced, and nothing is minted on the happy path.

## The user-facing API

`fetchEnv` now only fetches. No `env` option, no modes, no callbacks — it returns what the Neon API reports, and mints a credential only if you ask for a credential-backed var. It gains the same typesafe key filter `parseEnv` has:

```ts
const { storage } = await fetchEnv(config, {
    projectId,
    branch: "main",
    keys: ["AWS_ENDPOINT_URL_S3", "AWS_REGION"],
});
storage.endpoint; // string — `accessKeyId` is absent, and was never fetched
```

**The filter skips work, not just result fields.** Leave out `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `NEON_AI_GATEWAY_TOKEN` and no credential is minted at all — the endpoint, region and gateway host are branch metadata and come back on their own. That is the mechanism everything else is built on.

Everything stateful moved into one wrapper, outside `fetchEnv` and off the root entry point — it reads an env source and can mint and revoke credentials, so it ships from **`@neon/env/runtime`**:

```ts
import { fetchEnvReusingSecrets } from "@neon/env/runtime";

const { vars, credential } = await fetchEnvReusingSecrets(config, {
    projectId,
    branch: "main",
    env: { ...process.env, ...readEnvFile(".env") },
});

// vars: { DATABASE_URL: "…", AWS_ACCESS_KEY_ID: "…", … } — ready to write or inject
if (credential.issued) {
    console.log(`new values for ${credential.keys.join(", ")}`);
    // credential.revoked holds the token ids it superseded
}
```

It verifies the persisted secrets against the branch's live credentials, keeps them only when they name one that still exists, isn't revoked or expired, and carries every scope the policy needs — then asks `fetchEnv` for everything *except* what it kept, and revokes what the new credential supersedes. The report is a return value, so no callback is needed.

| Change | Effect on existing callers |
| --- | --- |
| `fetchEnv`'s `env` option removed | Callers that passed it move to `fetchEnvReusingSecrets` (all three in this repo did) |
| `fetchEnv` gains `keys` | Additive |
| `NEON_BRANCH` is now a selectable key | Additive; `parseEnv` filters can select it too |
| `toEntries` accepts a filtered result | Additive |
| New `@neon/env/runtime` subpath | Additive; the root export stays `fetchEnv` / `parseEnv` / `toEntries` / `NEON_ENV_VAR_KEYS` |

`neon env pull`, `neon dev`, and `neon-env run` / `export` all go through the wrapper, so **all three now verify rather than trust** — previously `dev` and `run` would happily inject a placeholder. New `env pull` output, only when a credential is actually re-issued:

```
Pulled 7 Neon variables into .env: NEON_BRANCH, DATABASE_URL, ..., AWS_ACCESS_KEY_ID, ...
Issued a new branch credential — these now hold fresh values: AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY
Revoked the credential it replaced (cred-fake-0001).
```

A verified reuse prints nothing extra: the "Pulled" line is now true for every key it lists, which is what #352 set out to fix by splitting the line.

## Revocation is narrow on purpose

Only credentials the persisted secrets *named*, and only those minted under this tool's own `neon-env <branch>` name. Their secrets lived nowhere but the env source being overwritten, so revoking them strands nothing. Everything else on the branch is left alone — it may belong to a teammate, another checkout, or a deployed function, and nothing observable tells those apart from an orphan of our own. A credential we didn't issue can still be *kept* if it fits; it just never gets revoked. Revocation happens *after* the fetch, so a failed fetch leaves the caller's existing secrets working.

Net effect: one credential per branch on the steady path, and the branch stops accumulating one per feature change.

## Also in here

- **`packages/ai-sdk-provider/.env.example`** — placeholders emptied. It was the only `.env.example` in the repo with non-empty values; the other five all use bare `KEY=`. An empty value was always falsy, so this alone fixes the reported symptom. Defence in depth, and it stops advertising a fake token shape.
- **Auth's env fallback moved with the rest.** `fetchEnv` used to fall back to its env source when an integration predating `base_url` returned an empty one. With no env source it returns `""`, and the wrapper keeps a non-empty persisted value over an empty fetched one — for every var, not just that one, so a pull can never blank a working line.
- **`fetchEnv` reads branch storage before minting.** A policy declaring buckets on a branch without storage used to mint a credential and *then* throw.
- **`FakeNeonApi` minted colliding `tokenIdShort` values** — `` `c${seq}`.slice(0, 12) `` truncated the digit that varies, so every credential shared one short id. Nothing looked credentials up by id before, so no test could catch it. This is what made my first draft of the mixed-halves test pass when it should have failed.
- **The root export stays pure.** `fetchEnvReusingSecrets` is the only stateful thing here, so it ships from a `@neon/env/runtime` subpath rather than the entry point an app or build script imports — the same split, for the same reason, as `@neon/config` vs `@neon/config-runtime`. Keeping it inside the package (rather than moving it into this CLI) means `neon-env run` / `export` keep verified reuse, and nothing has to re-derive which env vars a policy implies or which of them a credential backs. A contract test pins both surfaces.
- **`packages/cli/tsconfig.json` gains a `paths` entry** for the subpath: the CLI is on classic `moduleResolution: node`, which ignores package `exports`. Same treatment `neon-init/bootstrap` and `@neon/sdk/raw` already get.
- **`packages/env/CONTRIBUTING.md`** — the first per-package contributing file, documenting which entry point a change belongs in. The boundary is easy to erode by accident: when `packages/cli` needs an internal, the natural move is to widen the root export, which is the one thing the split exists to prevent. It records the test for which side code goes on, why reuse can't sit on the root export, why it stays in `@neon/env` rather than moving to the CLI, and the branch-credential invariants. Linked from the root `AGENTS.md`, the root `CONTRIBUTING.md`, and the package README.
- Changeset (`@neon/env` minor, `neon`/`neonctl` patch) and a rewritten README section.

## Verification

Branched off `d734cf2`. Every check CI runs, locally: `pnpm build`, `pnpm --filter neon typecheck`, `pnpm lint:ci`, `pnpm test:ci` — 3,938 tests green. 9 tests for the wrapper and 2 for `fetchEnv`'s filter, plus the 2 CLI pull tests, each written to fail first:

- a `.env.example` placeholder is replaced, and nothing is revoked (it named nothing)
- a live, sufficiently-scoped credential is kept — zero `createCredential`, and the non-secret storage vars still refresh
- a credential revoked out from under the env source is replaced
- enabling a second feature re-mints and revokes the credential it supersedes; the branch ends with exactly one
- a credential this tool didn't issue is never revoked
- storage and gateway halves naming different credentials re-mint (they share one credential, so disagreement means neither is trustworthy)
- a policy with neither feature never touches the credentials endpoint; a first run never calls `listCredentials`
- a persisted Auth base URL survives an integration that can no longer report one
- selecting only non-secret storage vars mints nothing; selecting a secret var does
- `env pull` twice in a row mints once and leaves the file byte-identical

**Also verified live** against a throwaway project in the Testing org (`org-old-flower-82714815`), with a real bucket on the branch — see the smoke-run comment below. `AWS_ACCESS_KEY_ID` came back byte-identical to the credential's `token_id`, the pulled credentials list the bucket over S3 with HTTP 200, a repeat pull mints nothing, a revoked credential is detected and replaced, and the scope-widening path revoked exactly the credential it superseded.

## For your attention

- **`neon dev` and `neon-env run` now cost one extra `listCredentials` call per start.** I argued against that in my review of #352, when verification bought nothing on those paths; now it comes bundled with not injecting placeholders, and it *avoids* the expensive call rather than adding one. A failure there degrades to "run without injection" on `dev`, as every other resolution failure does, and surfaces on `pull`.
- **The wrapper resolves the branch, then `fetchEnv` resolves it again** — two `listBranches` calls per pull. Threading a resolved snapshot through would remove one, at the cost of a private option on a function I just finished simplifying. Happy to add it if you'd rather.
- **Rotation is visible to anything else holding the old secrets.** If a value that fails verification was also pasted into a deployed preview, that copy stops working when the credential is revoked. It only happens for secrets that couldn't be verified in the first place, but it is a real behavior change.
- **Orphans from an overwritten or deleted `.env` are not collected — confirmed live.** In the smoke run, planting placeholders over a working credential left that credential live on the branch: nothing named it any more, so nothing revoked it. `neon-env <branch>` alone isn't proof it's mine rather than a teammate's, and I'd rather leak an occasional credential than revoke someone else's. `lastUsedAt` would let us guess, and I don't think we should guess.
- **`listCredentials` is not degraded.** If it fails on a branch that has storage or the gateway enabled, the pull fails. That endpoint has to work for `createCredential` to work anyway, and degrading it to `[]` would silently mint on every call — the spam this is designed to avoid.


andrelandgraf:
## Live smoke run

Ran the PR build (`packages/cli/dist/cli.js` at `0cedcbd`) against a real throwaway project in the Testing org — `smoke-env-pull-0ecc9a38` / `red-tooth-13934335`, `aws-us-east-2` — with a real bucket on `main` so the credential path was actually exercised. Deleted afterwards.

### The load-bearing claim, on real data

The whole design rests on the persisted secrets naming their own credential. After the first pull:

```
AWS_ACCESS_KEY_ID=nak_live_573cc63eeb504e7ea83c7ff962507cc9
```

```
GET /projects/red-tooth-13934335/branches/br-raspy-grass-axzxblef/credentials
→ nak_live_573cc63eeb504e7ea83c7ff962507cc9 | neon-env main | [storage:read, storage:write]
```

Byte-identical to `token_id`, named `neon-env main`, scoped from the policy. No local bookkeeping needed, confirmed against the live API rather than the fake.

### What each pull did

| Run | On disk beforehand | Output | Credentials on the branch |
|---|---|---|---|
| 1 — bucket just created | no `AWS_*` | `Pulled 7 …` + `Issued a new branch credential` | 1 live |
| 2 — nothing changed | valid credential | `Pulled 7 …`, no credential line | 1 live, file byte-identical |
| 3 — placeholders planted | `nak_live_...` / `your-secret-here` | `Pulled 7 …` + `Issued a new branch credential` | new one live, placeholder named nothing so nothing revoked |
| 4 — revoked key on disk | credential revoked server-side | `Pulled 7 …` + `Issued a new branch credential` | replaced |
| 5 — nothing changed | valid credential | no credential line | reused |

Run 2 and 5 are the anti-spam guarantee: a repeat pull mints nothing and leaves the file unchanged. Run 3 is the reported bug — the placeholder is gone, replaced by a working credential. Run 4 is the case presence-based reuse was equally blind to.

### The pulled credentials actually work

Signing a real S3 request with what `env pull` wrote, against the branch's own endpoint:

```bash
curl --aws-sigv4 "aws:amz:$AWS_REGION:s3" \
  --user "$AWS_ACCESS_KEY_ID:$AWS_SECRET_ACCESS_KEY" \
  "$AWS_ENDPOINT_URL_S3/smoke-assets/?list-type=2"
```

```
HTTP 200
<ListBucketResult><Name>smoke-assets</Name>…<KeyCount>0</KeyCount></ListBucketResult>
```

Still 200 after a pull that *reused* the credential, so verified reuse keeps a working one rather than merely an unchanged one. And this confirms the full token id is the right access-key id — the same request with the placeholder the old code would have silently kept:

```
HTTP 403
<Code>InvalidAccessKeyId</Code>
```

That 403 is the cost of the bug, measured.

### Revocation

The scope-widening path is the only one that revokes, so I drove `fetchEnvReusingSecrets` directly with a policy adding `aiGateway: true` on top of the existing bucket, against the same live branch:

```json
{
  "issued": true,
  "keys": ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "NEON_AI_GATEWAY_TOKEN"],
  "revoked": ["nak_live_37f795ceb31d456cae1d869148a606b2"]
}
```

Server-side afterwards:

```
nak_live_c035b7ba… | storage:read,storage:write,ai_gateway:invoke | revoked: no
nak_live_37f795ce… | storage:read,storage:write                  | revoked: 2026-07-28T15:35:09Z
```

Exactly one credential revoked — the one the persisted secrets named — and the replacement carries the widened scopes. The gateway base URL derived correctly too, cell prefix intact: `https://br-raspy-grass-axzxblef-api.ai.c-4.us-east-2.aws.neon.tech`.

### One thing this makes concrete

The branch ended with 3 live credentials and 1 revoked, across 5 pulls. That is the documented limitation, not a leak in the steady state: runs 3 and 4 each *replaced* a credential whose secrets had already been overwritten on disk, so nothing named the old one and nothing could safely attribute it. On the normal path — pull, pull, pull — it stays at exactly one, which runs 2 and 5 show. Worth knowing that planting placeholders over a working credential orphans it.

Nothing left behind: project deleted, run directory removed. OAuth session was enough throughout; no API key needed.

PR #362, opened 2026-07-31
feat(sdk): add a userAgent option so callers can identify themselves

andrelandgraf:
## Problem

Neon attributes API traffic to a caller by user agent — that is what makes CLI, MCP, and console traffic separable from direct API use in `fact_segment_api_events`. `@neon/sdk` sends no user agent, and `NeonConfig` accepts `apiKey`, `throwOnError`, `waitForReadiness`, `wait`, `retries`, `baseUrl`, `fetch`, and `orgId` — no `headers`, no `userAgent`. There is no supported way for an application built on this SDK to say who it is.

The consequence, from the shipped 1.4.0 artifact:

```
with userAgent : {"ua":"mcp-server-neon/1.0.0","auth":"Bearer test-key"}
without        : {"ua":"node","auth":"Bearer test-key"}
```

Every SDK consumer currently reports as `node` and lands in the undifferentiated `API` bucket.

This is not hypothetical. `mcp-server-neon` migrated to this SDK on 2026-07-22 and lost its attribution the same hour: `MCP` requests fell from 567,330/day to 19,957/day and distinct users from 14,815 to 587, while combined `MCP` + `API` users stayed flat (26,161 → 25,630). Its usage and its backend error rate — the only reliability signal that server has — have been measuring 4% of real traffic since. It is working around this gap by wrapping `fetch` ([mcp-server-neon#311](https://github.com/neondatabase/mcp-server-neon/pull/311)).

## The interface

```ts
const neon = createNeonClient({
  apiKey: process.env.NEON_API_KEY!,
  userAgent: "my-cli/1.2.0", // default: unset
});
```

| Option | Type | Default | Applies to |
|---|---|---|---|
| `userAgent` | `string` | — | `User-Agent` on every request the client makes |

It is set on the config passed to `createClient`, and `mergeHeaders(_config.headers, options.headers)` runs per request, so it covers the ergonomic namespaces and raw calls that reuse `neon.client` alike:

```ts
const neon = createNeonClient({ apiKey, userAgent: "my-cli/1.2.0" });

await neon.projects.get("p-1");                                  // sent
await getProject({ client: neon.client, path: { project_id } }); // sent
```

Nothing changes for existing callers. When `userAgent` is unset no header is added and requests carry the runtime default, exactly as today. Callers already passing a custom `fetch` are unaffected; a per-request header still wins over the client default, since `mergeHeaders` takes config first.

## Why an option rather than a default

The SDK could send `@neon/sdk/<version>` on its own. It deliberately doesn't, in this PR.

The control plane derives `api_source` from the user agent. Giving every existing SDK caller a new, uniform user agent changes how all of that traffic classifies, on the deploy of a patch release, with no way for a consumer to opt out of the reclassification. That is an analytics migration, and it wants to be decided with the data owners rather than shipped as a side effect of adding an option.

An opt-in option has no such effect: traffic only moves when a consumer chooses to identify itself, which is the point.

## Verification

Base `7ee4de5`. `pnpm --filter @neon/sdk build`, `test:ci` (37 passed), `test:types` (14 passed, no type errors), and `pnpm lint:ci` across 558 files all clean.

Four unit tests, each asserting on the `Request` the client actually constructed:

- an ergonomic call (`projects.get`) carries the configured user agent
- a raw call sharing `neon.client` carries it too
- unset adds no header at all, leaving the runtime default
- `Authorization` is unaffected

The first two were confirmed red before the change (`expected null to be 'my-cli/1.2.0'`); the last two pass either way and are there as controls.

Separately, the output quoted above came from `dist/index.js` after a real build, driven against a local HTTP server — the option survives bundling, which the source-level tests would not have caught.

Not verified against the live control plane: nothing exposes what user agent Neon recorded for a request, so the end-to-end claim rests on the header being on the wire. The observable confirmation will be the `MCP` series in `api_activity_daily` recovering once mcp-server-neon deploys.

## For your attention

- **`userAgent` is not accepted per call**, only on the client. Per-call overrides in this SDK are `{ throwOnError?, waitForReadiness?, signal? }`, and a caller's identity does not vary per request. Anyone who needs that can still pass `headers` on the individual raw call.
- **No validation on the string.** An invalid token would be rejected by the runtime's `fetch`, not here. Adding a format check would mean encoding RFC 9110 product-token rules for no case anyone has hit.
- **This does not fix existing consumers**, only unblocks them. `neonctl` and anything else in this monorepo built on the SDK still send `node` until each sets the option; whether they should is a separate change.

### DIFF
--- a/.changeset/olive-moons-shake.md
+++ b/.changeset/olive-moons-shake.md
@@ -1,0 +1,7 @@
+---
+"@neon/sdk": minor
+---
+
+Add a `userAgent` option so applications built on the SDK can identify themselves.
+
+Neon attributes API traffic to a caller by user agent, but the SDK sent none and its config exposed no way to set one, so every SDK-backed tool was indistinguishable from a direct API call. `createNeonClient({ apiKey, userAgent: "my-cli/1.2.0" })` now sets the header on every request the client makes, including raw calls that reuse `neon.client`. Unset by default, so nothing changes for existing callers.
--- a/packages/sdk/README.md
+++ b/packages/sdk/README.md
@@ -45,9 +45,23 @@ const { project, connectionString } = data;
 | `retries` | `number` | `2` | Automatic retries on always-safe statuses (`423`, `429`, `503`) with backoff. |
 | `orgId` | `string` | — | Default organization, applied to project create/list and as the transfer source org. Overridable per call. |
 | `baseUrl` | `string` | `https://console.neon.tech/api/v2` | Override the API base URL. |
+| `userAgent` | `string` | — | Product token sent as `User-Agent` on every request, e.g. `my-cli/1.2.0`. See below. |
 | `fetch` | `typeof fetch` | global `fetch` | Custom fetch implementation (proxies, tests, non-global runtimes). |
 
 Every option except `apiKey` is also accepted **per call** via the last `options` argument (`{ throwOnError?, waitForReadiness?, signal? }`), overriding the client default.
+
+### Identifying your application
+
+Neon attributes API traffic to a caller by user agent. If you are building a tool, integration, or service on this SDK and want its usage counted separately from direct API calls, set `userAgent`:
+
+```ts
+const neon = createNeonClient({
+  apiKey: process.env.NEON_API_KEY!,
+  userAgent: "my-cli/1.2.0",
+});
+```
+
+It applies to every request the client makes, including raw calls that reuse `neon.client`. Left unset, requests carry whatever your runtime's `fetch` sends by default — `node` on Node.js — and are indistinguishable from any other client.
 
 ## The result model
 
--- a/packages/sdk/src/neon/config.ts
+++ b/packages/sdk/src/neon/config.ts
@@ -28,6 +28,14 @@ export interface NeonConfig<Throw extends boolean = false> {
 	retries?: number;
 	/** Override the API base URL. Defaults to `https://console.neon.tech/api/v2`. */
 	baseUrl?: string;
+	/**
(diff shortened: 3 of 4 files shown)


andrelandgraf:
Closing without merging. The option is not worth the surface.

`fetch` already covers this. A wrapper that adds one header is four lines, composes with everything else the SDK does, and needs no new config:

```ts
const neon = createNeonClient({
  apiKey,
  fetch: (input, init) => {
    const request = new Request(input, init);
    request.headers.set("User-Agent", "my-tool/1.2.0");
    return fetch(request);
  },
});
```

The problem statement in this PR claimed there was no supported way for an application to identify itself. That was wrong, and [mcp-server-neon#311](https://github.com/neondatabase/mcp-server-neon/pull/311) is shipping on the wrapper.

What was left after that correction was a discoverability argument: a `userAgent` row in the config table would tell consumers that attribution exists and is expected. That reasoning holds for Neon's own tools, which is where the demand actually comes from — we want our SDK-backed surfaces tagged so we can measure them. It does not generalize to the people this SDK is published for, who have no reason to want their traffic labelled. Adding a permanent public field to serve an internal need is the wrong trade.

One thing worth carrying over for anyone else wrapping `fetch` around this client: the SDK calls it as `_fetch(request)` with a fully-built `Request` and no second argument. Building headers from `init` and passing them back produces this:

```
naive wrapper    {"ua":"my-cli/1.0.0"}                        <- Authorization gone
correct wrapper  {"ua":"my-cli/1.0.0","auth":"Bearer k"}
```

`init.headers` replaces rather than merges, so the auth header is dropped and every call 401s. Construct a `Request` and set the header on it.

PR #387, opened 2026-08-05
Make externalPackages ship a package's files, with includeFiles to opt out

monsterzz:
Supersedes the `nativePackages` design this branch previously carried. The tracer, the
schema discipline, and the reasoning about archive layout are Gleb's; what changed is the
shape of the option users write.

## The problem

`externalPackages` shipped in `@neon/config@0.11.0` and is live on npm today. It externalizes
an import and ships nothing, so the package throws `Cannot find module` if the function ever
reaches it. Its own JSDoc on `main` spends twenty lines saying so:

> **An external package is not resolvable at runtime.** […] This option therefore only
> unblocks an import that is never evaluated; it does not make a dependency usable.
>
> […] such a package cannot work on Functions until the deployed archive can carry files
> alongside the bundle. Do not reach for `externalPackages` to try: it moves the error from
> deploy to invoke.

An option whose documentation is mostly a warning against using it has the wrong default.

The name is the load-bearing part. `externalPackages` is what a user guesses, because
Next.js's `serverExternalPackages` means *externalize and it still works* — Vercel traces the
package's files into the deployment automatically, so externalizing implies shipping. A user
carrying that model reaches for the name they know and silently gets the broken-at-invoke
behaviour, while the working one hides behind a second name they have to be told about.

## What changed

One list, with a mode. `nativePackages` is gone.

```ts
export default defineConfig({
  preview: {
    functions: {
      resize: {
        name: "Resize",
        source: "./functions/resize.ts",
        externalPackages: [
          "sharp",                                  // unbundled, files ship — it works
          { name: "canvas", includeFiles: false },  // unbundled, nothing ships
        ],
      },
    },
  },
});
```

| Form | Meaning |
| --- | --- |
| `"sharp"` | externalized, the package's real files are staged into the archive |
| `{ name, includeFiles: true }` | identical to the bare string |
| `{ name, includeFiles: false }` | externalized, nothing staged — the previous behaviour |

`includeFiles` is the noun Vercel already uses for this operation (`functions.includeFiles`
in `vercel.json`). Theirs takes a glob and ours takes a boolean on a package entry, which is
the one wrinkle in borrowing the name.

**Why one list and not two.** The two options were one concept with a mode, and the tell was
the cross-list validation rule the previous design needed: a rule that exists only to reject
an incoherent combination means the split is artificial. Nothing validates that `env` and
`dev` do not overlap. With a union entry the contradiction is unrepresentable.

## Types

`ResolvedFunctionConfig.externalPackages` changes from `string[]` to a normalized shape, so
nothing downstream re-derives the default:

```ts
interface ResolvedExternalPackage {
  name: string;
  includeFiles: boolean;
}

externalPackages?: ResolvedExternalPackage[];
```

Two pure helpers are exported for anyone implementing a custom `FunctionBundler`, which is a
documented extension point that receives a resolved config:

```ts
import { packagesToStage, externalPackageRoot } from "@neon/config";

packagesToStage([{ name: "sharp", includeFiles: true }]);        // ["sharp"]
externalPackageRoot("@scope/pkg/sub/deep.js");                    // "@scope/pkg"
```

## Validation

Subpaths stay legal — esbuild accepts them, and the specifier is passed through to the trace
as authored, because a package may export only a subpath and be unimportable by its root.
Three new rules replace the cross-list overlap check:

```
externalPackages[1]: "sharp" is listed more than once (first at index 0)

externalPackages[1]: "sharp/lib/index.js" and "sharp" are both part of the "sharp" package
but disagree about includeFiles; files ship per package, so the whole package either ships
or does not

externalPackages[0]: "@scope/*" does not name a single installable package, so its files
cannot be staged. Name one package, or set includeFiles: false to leave the import
unresolved without shipping anything for it
```

That last rule is new exposure created by the flip: esbuild's `external` accepts a `*`
wildcard and a bare scope, which name a set rather than a package. Before, nothing was
installed for them. Now a staged entry's root is handed to `npm install`, so it has to name
one package — unless it stages nothing, where the wildcard forms stay legal.

## What does not change

A function that declares no `externalPackages` resolves to exactly the shape it did before —
the property is absent rather than `[]`, so the bundler's pre-existing path is untouched and
the archive is byte-identical. Same for a function whose every entry sets
`includeFiles: false`.

## Also in here

- The in-repo readers of `ResolvedFunctionConfig.externalPackages` (`neonctlBundler`,
  `neon dev`'s function planner, `buildFunctionBundle`'s esbuild call) map to names, so the
  repo stays green on this commit alone. Their behaviour is unchanged: everything is
  externalized exactly as before, and nothing is staged until the bundler half lands.
- Rebased onto `main`, which had moved 9 commits ahead.

## Verification

`pnpm build`, `pnpm lint:ci`, and `pnpm test:ci` are green across the workspace —
`@neon/config` at 257 tests, 4470 across all packages.

New coverage, as behaviours:

- a bare string resolves to `includeFiles: true`; the object form with and without the flag;
  an explicit `true` is identical to the bare string
- an undeclared policy resolves with the property absent, not empty
- the policy array is copied, so mutating it afterwards cannot reach the resolved config
- the object form rejects an unknown key, a non-boolean `includeFiles`, a missing `name`, and
  a relative path
- a subpath entry is accepted; a duplicate name is rejected against the offending index; a
  bare name and a subpath of it are rejected when they disagree and accepted when they agree;
  two packages sharing a scope are not treated as one
- `externalPackageRoot` across bare, scoped, subpath, and scoped-subpath specifiers
- `packagesToStage` drops opted-out entries, keeps a subpath specifier as authored, and
  deduplicates an exactly repeated one
- a wildcard, a bare scope, and a protocol specifier are rejected when staged and accepted
  with `includeFiles: false`

## For your attention

- **This changes behaviour that is already published.** `@neon/config` 0.11.0–0.14.0 are on
  npm with the old meaning, so `externalPackages: ["x"]` today ships nothing and after this
  stages `x`. The exposure is four days of a preview-only option on a 0.x package, and the
  break is loud rather than silent: a package with no linux-arm64 build fails the deploy with
  a message naming `includeFiles: false` as the fix. A reviewer who thinks that is the wrong
  trade should say so now, because it is much more expensive to reverse after another release.
- **`ResolvedFunctionConfig.externalPackages` is a public type.** Anything reading a resolved
  config or implementing `FunctionBundler` sees the shape change. `packagesToStage` exists so
  those callers do not hand-roll the default.
- The subpath rule is looser than the previous `nativePackages` rule, which rejected subpaths
  outright. Rejecting them here would have broken published configs that use one.

### DIFF
--- a/.changeset/tidy-moons-gather.md
+++ b/.changeset/tidy-moons-gather.md
@@ -1,0 +1,9 @@
+---
+"@neon/config": minor
+---
+
+Add `nativePackages` to a `neon.ts` function definition, declaring packages backed by a
+native binary whose real files should ship into the deployed archive alongside the bundle.
+This is the schema and type surface only; the bundler change that acts on the list follows
+separately. A package may not appear in both `nativePackages` and `externalPackages`, and
+entries are package names without a subpath.
--- a/.changeset/tidy-moons-gather.md
+++ b/.changeset/tidy-moons-gather.md
@@ -2,8 +2,31 @@
 "@neon/config": minor
 ---
 
-Add `nativePackages` to a `neon.ts` function definition, declaring packages backed by a
-native binary whose real files should ship into the deployed archive alongside the bundle.
-This is the schema and type surface only; the bundler change that acts on the list follows
-separately. A package may not appear in both `nativePackages` and `externalPackages`, and
-entries are package names without a subpath.
+**Breaking:** `externalPackages` now ships a package's real files into the deployed archive
+by default, instead of externalizing the import and shipping nothing.
+
+A bare string is the common form and produces a function that works:
+
+```ts
+externalPackages: ["sharp"],
+```
+
+The previous behaviour — externalize the import, ship nothing, throw `Cannot find module`
+if the function reaches it — is now the opt-out, for a package that cannot be staged and is
+never actually reached:
+
+```ts
+externalPackages: ["sharp", { name: "canvas", includeFiles: false }],
+```
+
+`ResolvedFunctionConfig.externalPackages` changes from `string[]` to
+`{ name: string; includeFiles: boolean }[]`, which affects anything reading a resolved
+config or implementing a custom `FunctionBundler`. Two pure helpers are exported for that
+case: `packagesToStage` and `externalPackageRoot`.
+
+A function that declares no `externalPackages`, or whose every entry sets
+`includeFiles: false`, deploys exactly the archive it did before.
+
+Validation additionally rejects the same package listed twice, and a bare name and a subpath
+of it that disagree about `includeFiles`. This is the schema and type surface; the bundler
+change that stages the files lands separately.
--- a/packages/cli/src/commands/config.ts
+++ b/packages/cli/src/commands/config.ts
@@ -69,7 +69,11 @@ const neonctlBundler: FunctionBundler = async (fn) =>
 	zipBundle(
 		await bundleEntry(fn.source, {
 			...(fn.externalPackages
-				? { externalPackages: fn.externalPackages }
+				? {
+						externalPackages: fn.externalPackages.map(
+							(pkg) => pkg.name,
+						),
+					}
 				: {}),
 		}),
 	);
--- a/packages/cli/src/dev/functions.ts
+++ b/packages/cli/src/dev/functions.ts
@@ -79,8 +79,14 @@ export const resolveFunctionsFromConfig = async (
 				? { port: devPort(fn.dev) as number }
 				: {}),
 			env: { ...fn.env },
+			// Names only: locally every entry is simply left unbundled, and `includeFiles`
+			// governs the deployed archive, which `neon dev` does not build.
 			...(fn.externalPackages
-				? { externalPackages: [...fn.externalPackages] }
+				? {
+						externalPackages: fn.externalPackages.map(
+							(pkg) => pkg.name,
+						),
+					}
 				: {}),
 		};
 	});
--- a/packages/config-runtime/src/lib/function-bundle.test.ts
+++ b/packages/config-runtime/src/lib/function-bundle.test.ts
@@ -24,7 +24,14 @@ function fn(
 		source,
 		env: {},
 		runtime: "nodejs24",
-		...(externalPackages ? { externalPackages } : {}),
+		...(externalPackages
+			? {
+					externalPackages: externalPackages.map((name) => ({
+						name,
+						includeFiles: true,
+					})),
+				}
+			: {}),
 	};
 }
 
--- a/packages/config-runtime/src/lib/function-bundle.ts
+++ b/packages/config-runtime/src/lib/function-bundle.ts
@@ -74,11 +74,10 @@ export async function buildFunctionBundle(
 			// `platform: "node"`. The banner re-creates require/__filename/__dirname so
 			// bundled CommonJS deps work inside the ESM output.
 			banner: { js: ESM_CJS_INTEROP_BANNER },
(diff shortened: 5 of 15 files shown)


monsterzz:
@andrelandgraf for review when you have a moment. This is the schema half; the bundler half is the PR titled "Ship a function's native package files into the deployed archive" (#388), which should merge after this one.

andrelandgraf:
Andre's agent: Andre asked me to rework the option shape on both of your PRs, so I pushed to
these branches rather than opening competing ones. Your commit is preserved as the parent on
this branch, and you are co-author on the bundler commit in #388. Both were 9 commits behind
`main`, so they are rebased — you will need to reset your local copies.

What changed and why:

`nativePackages` is merged into `externalPackages` as `includeFiles`, defaulting to true. The
argument is the name: `externalPackages` is what a user guesses, because Next.js's
`serverExternalPackages` means externalize-and-it-still-works, so the guessable name should
do the working thing and the broken-if-reached state should need the explicit gesture. The
overlap-validation rule your version needed was the signal — a rule that exists only to
reject an incoherent combination usually means the two lists are one option with a mode.

Everything mechanical in #388 is yours and unchanged: the cross-install, the nft trace, the
variant filter, the ELF check, the layout preservation, the size limits. I added detection
that reports a bundled-in native package nobody declared, since `sharp` produces no build
error and nothing else catches it before production.

Three things I changed that are corrections rather than redesign, and you may disagree:

1. Version pinning never worked for `sharp`. `require.resolve("sharp/package.json")` throws
   `ERR_PACKAGE_PATH_NOT_EXPORTED` because its `exports` map lists only `.`, so the deploy
   was staging the registry's latest. Now read off disk. There is a regression test with that
   exact manifest shape.
2. The `neon function deploy --src` remediation in two error messages is not true — that path
   zips only the esbuild output and cannot carry extra files.
3. The live install tests no longer skip. CI's `prepare` writes an `~/.npmrc` pointing at the
   Databricks JFrog mirror before anything installs, and a spawned `npm` reads it, so the
   no-egress reason does not hold. If CI proves me wrong I will re-gate them.

Detection is a warning and never fails a deploy. `ws` with `bufferutil`, or a Prisma client
with no Rust engine, carry native code on a path they never reach and deploy fine today;
failing on that evidence would break them.

If the default flip is the wrong call for this option, say so — it is much cheaper to
reverse now than after another release.

andrelandgraf:
Andre's agent: closing this in favour of #398, which is the same change on a branch in this
repo rather than on your fork. Not a rejection of the work — it is a CI constraint I only
found after pushing to your branch.

CI cannot run on a fork PR here. `ci.yml` declares `id-token: write`, but GitHub does not
issue an OIDC token to a `pull_request` workflow from a fork, and
`./.github/actions/prepare` needs one to authenticate against the Databricks JFrog mirror
before anything installs. So every job dies before `pnpm install`:

```
Failed to fetch OpenID Connect JSON Web Token:
Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable
```

The checks on this PR had been sitting at `action_required` since 5 August, so it was never
verified by CI at all — that predates my changes. A PR's head repository cannot be changed
after it is opened, which is why this needs a new PR rather than a fix here.

#398 and #399 carry your commit and your co-authorship, and the tracer pipeline is unchanged
in substance. The option shape changed — `nativePackages` merged into `externalPackages` as
`includeFiles`, defaulting to true — and the reasoning for that is in the comment above and
in #398's description. If the default flip is the wrong call, that is worth arguing on #398
before it merges.

Worth someone deciding separately whether fork PRs should be able to build at all in this
repo, since right now an outside contributor gets no CI signal.

PR #388, opened 2026-08-05
Stage external package files, and report an undeclared native dependency

monsterzz:
Depends on #387, which merges `nativePackages` into `externalPackages`. **Merge that first.**

The install / trace / verify pipeline here is Gleb's and is unchanged in substance. This
branch adapts it to the merged option and adds the reporting half.

## The problem

A package backed by a `.node` binary could not work on Functions. The binary is a compiled
object the platform loads from a real path, so no bundler can inline it, and the archive was
a single `index.mjs` with no `node_modules` beside it for an externalized import to resolve
against.

`sharp` is the sharp edge. It loads its binary through `createRequire`, which esbuild does
not follow, so it bundles cleanly at ~40 KB and then dies at invoke:

```
Could not load the "sharp" module using the linux-arm64 runtime
```

**The runtime was never the obstacle.** It imports the entry by absolute path, so Node
resolves `node_modules` next to it — for both `import` and the `createRequire` shim the
bundle's banner already installs — from any working directory. The archive simply never
carried a `node_modules` to resolve against. So this is a bundler and packaging change; no
dataplane change was needed.

## Staging

Three steps, in an order that is not interchangeable.

**1. Install for the runtime target, not the deploy machine.**
`npm install <pkg> --cpu=arm64 --os=linux --libc=glibc` into a throwaway directory.

The user's own `node_modules` is never read for these files. Its binaries are built for their
machine, and — verified — a cross-platform install is **not sticky**: a later plain
`npm install` re-resolves optional dependencies for the host and silently replaces the
target's packages with the host's. `--ignore-scripts` is not only for speed; an install
script would compile for the host and overwrite the prebuilt binary being collected. npm is
spawned rather than imported, because the standalone CLI build contains neither npm nor a
`node_modules`.

**2. Trace the installed tree with `@vercel/nft`.** The file set cannot be discovered by
asking Node. `@img/sharp-linux-arm64` exports only `./sharp.node` and `./package` — no `.`,
no wildcard — so its directory is unreachable through the resolver. A tracer that follows
`createRequire` is the only thing that finds these files, which is the same blindness that
makes esbuild miss them in the first place. Sibling platform builds are then dropped, identified from their own manifests the way npm
decides what to install — `@img/sharp-linuxmusl-arm64` declares `libc: ["musl"]`, which a
glibc runtime cannot load. The WebAssembly fallback is the exception: `@img/sharp-wasm32`
declares no `os`, `cpu`, or `libc` at all, so only its name identifies it, matched as a whole
dash-delimited token. Dropping it is not a size saving — sharp's loader falls through to it
silently, so a broken native path would succeed on much slower code instead of failing
loudly.

**3. Copy the traced files preserving the `node_modules` layout.** A `.node` addon locates
its sibling shared libraries by a path relative to its own directory:

```
$ readelf -d sharp-linux-arm64-0.35.3.node
(NEEDED) libvips-cpp.so.8.18.3
(RPATH)  $ORIGIN/../../sharp-libvips-linux-arm64/lib:...
```

Flattening the tree breaks loading with every file present, so archive keys carry their full
`node_modules/...` path and a test asserts nothing collapses to a basename.

Failures are reported at deploy rather than at invoke: no build for the target; npm's
`EBADPLATFORM`, whose own message compares against the *host* and so reads as though the
wrong thing was asked for; a binary that is not an aarch64 ELF; a symlink, which the archive
cannot represent; and the size and entry limits, whose message names the largest
contributors.

## Reporting an undeclared native dependency

Staging only helps a user who knows the option exists. `sharp` gives no build error to hang
a message off — it bundles cleanly and fails in production — so nothing catches the common
case today.

esbuild's metafile names every package that was bundled in, and the path it reports is also
where that exact copy lives — which is what makes this work under pnpm, whose transitive
dependencies are not linked at the project root and whose store holds several versions of a
package side by side. For each package not already declared, its own metadata is checked for
native code: a `.node`/`.so` in its tree,
a platform-gated optional dependency carrying one (the `sharp` → `@img/sharp-linux-arm64`
shape, which scanning only the parent misses), a node-gyp install script, or a node-pre-gyp
`binary` field. A package whose `os`/`cpu`/`libc` exclude the runtime is skipped — `fsevents` is darwin-only
and a musl build cannot load on glibc, so neither is shippable and reporting them would be
noise on every macOS project. Manifests are read defensively: `os` may be a bare string, and
a malformed value must not take an advisory scan, and with it the deploy, down.

Real output, from a function importing `sharp` with nothing declared:

```
Function "resize" bundles "sharp", and it loads a compiled binary from
@img/sharp-darwin-arm64 (lib/sharp-darwin-arm64-0.35.3.node).
A compiled binary cannot be bundled, so if this function reaches that code path it
will fail at invoke rather than at deploy.

  Ship its files with the deploy:
      externalPackages: ["sharp"]

  Or, if this function never reaches it, silence this by saying so:
      externalPackages: [{ name: "sharp", includeFiles: false }]

Shipping a package adds its whole file tree to the archive, which increases both
archive size and cold start. If "sharp" is only used behind a JavaScript fallback,
the deploy is already correct and the second form records that.
```

Findings are returned, never printed: `pushConfig` puts them on `PushResult.warnings` and
the CLI renders them. A library has no business choosing an output channel, and the repo's
`noConsole` rule agrees.

**It is a warning, never a failure, and that is the central design decision here.** The
evidence proves the package contains compiled code; it cannot prove this function's code path
reaches it. `ws` with `bufferutil` installed, a Prisma client configured with no Rust engine,
and any package with an optional native accelerator behind a working JavaScript fallback all
look identical from here and deploy correctly today. Failing on that evidence would break
working deploys, so the message states the failure conditionally and names the
no-change-needed case.

Two consequences worth knowing:

- **It only runs after a successful bundle.** A package esbuild cannot resolve, or a `.node`
  it has no loader for, throws before any metafile exists — those already fail loudly with
  esbuild's own diagnostic. The case this covers is precisely the one that otherwise reaches
  production.
- **It also runs under `neon dev`**, as a warning. Locally these packages resolve from the
  real `node_modules` at host architecture, so a native dependency works perfectly on your
  laptop and gives no signal at all. Without this the first news is a failed deploy.

## `neon dev`

The declared names are mirrored into esbuild's `external` and nothing else happens: no
install, no copy. The dev bundle is written *inside* the project's own `node_modules`, so an
unbundled package resolves from the real tree at the host architecture. That was previously
incidental to where the bundle directory happened to sit; it is now a named helper with tests
pinning both the path and the resolution behaviour, because moving that directory would break
local dev for native dependencies and nothing else would catch it.

## Also in here

- **Version pinning was broken for the canonical package.** `installedVersion` used
  `require.resolve("<pkg>/package.json")`, which respects the `exports` map. `sharp` publishes
  `exports: { "." : … }` with no `./package.json`, so the lookup threw
  `ERR_PACKAGE_PATH_NOT_EXPORTED` even with `sharp` installed, and the deploy silently staged
  the registry's latest instead of the user's version. Found by the new warning below firing
  on a real project. Versions are now read from the package directory, located by walking
  `node_modules` the way Node does.
- **`NativeTraceResult.warnings` was always `[]`.** It now carries the case where a version
  cannot be determined, and both bundlers surface it, so staging a different version than the
  one you tested is said out loud instead of discovered from a deployed archive.
- **The trace imports the specifier as authored, while the install gets the package root.** A
  package may export only a subpath — `exports: { "./native": … }` with no `.` — in which
  case importing the root throws `ERR_PACKAGE_PATH_NOT_EXPORTED` and the trace finds nothing.
- **A traced file that is not in the staging tree now fails the deploy.** It was skipped,
  which produces an archive that is quietly missing a file and fails at invoke.
- **The `wasm32`/`musl` variant filter matches package directory names**, not the whole path.
  As a substring test it also deleted ordinary files such as `lib/musl-helper.js`.
- **Two deploy messages pointed at `neon function deploy --src`** as the way to ship a
  hand-built binary. That command runs `bundleEntry` and zips only its output — it cannot
  carry extra files, so the advice was false. Replaced with what is actually true.
- **The live install tests no longer skip.** They were gated behind
  `NEON_TEST_NATIVE_INSTALL=1` because the protected runners have no public egress, but CI's
  `prepare` action writes an `~/.npmrc` pointing at the Databricks JFrog mirror before
  anything installs, and a spawned `npm` reads it. They run in ~6s locally and really do
  install `sharp` for linux-arm64 and assert the AArch64 ELF. **If CI disagrees, that is a
  fact I could not check from here and I will re-gate them.**
- `bundleEntry` returns `{ files, metafile }` instead of a bare file map. CLI-internal, three
  call sites.
- `@vercel/nft` is a new dependency of `@neon/config-runtime`, loaded by dynamic `import()`
  like `esbuild` and `fflate` already are, so a `neon.ts` that only reads config never pulls
  it in.

## Verification

`pnpm build`, `pnpm lint:ci`, and `pnpm test:ci` are green across the workspace: 4488 tests.
`@neon/config-runtime` has no skipped tests; the CLI's 6 skips and `ai-sdk-provider`'s 1 are
pre-existing on `main` and untouched here.

Unit coverage runs against stubbed install and trace steps, so it needs no registry: the
unflattened layout, both size limits, the wrong-architecture / non-ELF / too-short-to-be-a-
binary rejections and the accepted aarch64 case, the manifest-driven sibling filter (and that
a package merely named `musl-helper` keeps its files), version pinning (including the
`exports`-map shape that broke it, and a scoped package), the package-root-versus-authored-
specifier split, and the never-read-the-user's-tree property. `pushConfig` is covered against
a real on-disk `node_modules` for both the reported and the nothing-to-report case. Detection is tested against
real `node_modules` trees written to temp directories rather than a faked filesystem, because
the thing under test is how a package looks on disk.

End to end against a real `sharp` install, all three states of the option:

| `externalPackages` | archive | reported |
| --- | --- | --- |
| absent | 1 entry, 0.04 MiB | the warning above |
| `["sharp"]` | 89 entries, 8.05 MiB, both AArch64 binaries present as siblings under `node_modules/@img` | nothing |
| `[{ name: "sharp", includeFiles: false }]` | 1 entry | nothing |

**Not verified: executing the arm64 binary end to end.** That needs a real deploy; this host
is x86_64, where loading it fails as expected. Everything up to and including the bytes that
would ship is verified.

## For your attention

- **The warning will fire on projects that are already correct.** `ws` with `bufferutil`
  installed is the common one. The message says so and `includeFiles: false` records the
  intent, but this is noise added to deploys that do not need changing, and the false-positive
  rate is the thing to push back on if it feels wrong.
- **Archive size.** `sharp` alone is 8.05 MiB compressed, which fits current limits with room
  to spare, but a couple of native packages is the practical ceiling. The size check reads its
  limits from an argument rather than hardcoding them.
- **Cold start** grows for functions that stage anything — 1 file to ~89, read file by file
  when a VM loads. Confined to functions that declare it, but worth measuring before this is
  recommended broadly.
- **Detection cannot see through Yarn PnP**, which resolves out of zip archives with no
  `node_modules` segment, and misses natives with no `.node`/`.so` (a Prisma engine, an
  `ffmpeg` binary) and anything loaded through a computed `require`. A missed package is a
  missing warning rather than a wrong one, which is the safe direction, but it means this is
  a net that catches the common case, not a guarantee.
- **Staging does not reproduce the user's lockfile.** It installs `<name>@<version>` for the
  declared package into a clean directory, so the root version is the one they resolved but
  transitive pins, overrides, patches, and aliases are not carried over. Making this exact
  means reading each package manager's lockfile, which is its own piece of work. A package
  whose version cannot be read at all is now reported rather than silently taking the
  registry's latest, which was the previous behaviour and was hitting `sharp`.
- **A package that is itself platform-gated cannot be staged directly.** Declaring
  `@img/sharp-linux-arm64` rather than `sharp` fails with `EBADPLATFORM`, because npm
  evaluates the root package against the host even when the target flags are passed. Declare
  the parent package instead. The error message names the target rather than explaining this,
  which is a rough edge.
- **`PushResult` gains a required `warnings` field.** Additive for anyone reading a result;
  breaking for anything constructing one.
- **The tracer's own warnings are deliberately not surfaced, and I tried the opposite
  first.** Reporting nft's unresolved-dependency warnings sounds right and measures wrong on
  the canonical package: tracing a correctly staged `sharp` emitted twelve of them, one per
  platform that was correctly not installed plus several for specifiers built by string
  concatenation that no tracer can follow, and none indicated a missing file. A report that
  is wrong every time trains people to ignore it. An archive that really is incomplete is
  caught by the checks that prove it instead — a traced file absent from staging fails the
  deploy, and a shipped binary that is not an AArch64 ELF fails it too.
- **A merged changeset is not a published package** — this needs a release before the option
  works from an installed CLI.

### DIFF
--- a/.changeset/lucky-rivers-travel.md
+++ b/.changeset/lucky-rivers-travel.md
@@ -1,0 +1,26 @@
+---
+"@neon/config-runtime": minor
+"neon": minor
+---
+
+Stage the real files of a function's `externalPackages` into the deployed archive, so a
+package backed by a native binary works on Functions instead of failing at invoke. Each
+declared package is installed for the runtime target (linux-arm64, glibc) into a throwaway
+directory, traced for the files it actually reaches, and copied under `node_modules/` with
+its directory layout preserved. The user's own `node_modules` is never read for those files
+or modified.
+
+An entry with `includeFiles: false` is externalized and nothing is staged for it, which is
+the pre-existing behaviour. A function whose entries all opt out — or which declares none —
+produces a byte-identical archive to before.
+
+Deploys and `neon dev` now report a package that was bundled in, carries native code, and
+was never declared. The report is advisory and never fails a deploy: the evidence shows the
+package contains compiled code, not that this function reaches it, and a package with a
+working JavaScript fallback looks identical.
+
+Fixes version pinning for packages whose `exports` map does not list `./package.json`.
+`sharp` is one, so the version the user had installed was never read and the registry's
+latest was staged instead. Versions are now read from the package directory rather than
+through the resolver, and a package whose version still cannot be determined is reported
+instead of being staged silently.
--- a/.changeset/tidy-moons-gather.md
+++ b/.changeset/tidy-moons-gather.md
@@ -1,0 +1,9 @@
+---
+"@neon/config": minor
+---
+
+Add `nativePackages` to a `neon.ts` function definition, declaring packages backed by a
+native binary whose real files should ship into the deployed archive alongside the bundle.
+This is the schema and type surface only; the bundler change that acts on the list follows
+separately. A package may not appear in both `nativePackages` and `externalPackages`, and
+entries are package names without a subpath.
--- a/.changeset/tidy-moons-gather.md
+++ b/.changeset/tidy-moons-gather.md
@@ -2,8 +2,31 @@
 "@neon/config": minor
 ---
 
-Add `nativePackages` to a `neon.ts` function definition, declaring packages backed by a
-native binary whose real files should ship into the deployed archive alongside the bundle.
-This is the schema and type surface only; the bundler change that acts on the list follows
-separately. A package may not appear in both `nativePackages` and `externalPackages`, and
-entries are package names without a subpath.
+**Breaking:** `externalPackages` now ships a package's real files into the deployed archive
+by default, instead of externalizing the import and shipping nothing.
+
+A bare string is the common form and produces a function that works:
+
+```ts
+externalPackages: ["sharp"],
+```
+
+The previous behaviour — externalize the import, ship nothing, throw `Cannot find module`
+if the function reaches it — is now the opt-out, for a package that cannot be staged and is
+never actually reached:
+
+```ts
+externalPackages: ["sharp", { name: "canvas", includeFiles: false }],
+```
+
+`ResolvedFunctionConfig.externalPackages` changes from `string[]` to
+`{ name: string; includeFiles: boolean }[]`, which affects anything reading a resolved
+config or implementing a custom `FunctionBundler`. Two pure helpers are exported for that
+case: `packagesToStage` and `externalPackageRoot`.
+
+A function that declares no `externalPackages`, or whose every entry sets
+`includeFiles: false`, deploys exactly the archive it did before.
+
+Validation additionally rejects the same package listed twice, and a bare name and a subpath
+of it that disagree about `includeFiles`. This is the schema and type surface; the bundler
+change that stages the files lands separately.
--- a/packages/cli/src/commands/config.ts
+++ b/packages/cli/src/commands/config.ts
@@ -71,9 +71,10 @@ import { autoPullEnvAfterPin } from "./env.js";
  */
 const neonctlBundler: FunctionBundler = async (fn) => {
 	const externalPackages = fn.externalPackages ?? [];
-	const { files, metafile } = await bundleEntry(fn.source, {
+	const { files, metafile, warnings } = await bundleEntry(fn.source, {
 		externalPackages: externalPackages.map((pkg) => pkg.name),
 	});
+	for (const warning of warnings) log.warning(warning);
 
 	// Advisory only — the evidence cannot prove the code path is reached, so a package with a
 	// working JavaScript fallback must not have its deploy blocked. See native-detect.
--- a/packages/cli/src/commands/config.ts
+++ b/packages/cli/src/commands/config.ts
@@ -69,7 +69,11 @@ const neonctlBundler: FunctionBundler = async (fn) =>
 	zipBundle(
 		await bundleEntry(fn.source, {
 			...(fn.externalPackages
-				? { externalPackages: fn.externalPackages }
+				? {
+						externalPackages: fn.externalPackages.map(
+							(pkg) => pkg.name,
+						),
+					}
 				: {}),
 		}),
 	);
--- a/packages/cli/src/commands/config.ts
+++ b/packages/cli/src/commands/config.ts
@@ -1,12 +1,15 @@
 import { existsSync, readFileSync, writeFileSync } from "node:fs";
-import { join } from "node:path";
-import { resolveConfig } from "@neon/config";
+import { dirname, join } from "node:path";
+import { packagesToStage, resolveConfig } from "@neon/config";
 import {
 	apply,
+	assertZipWithinLimits,
 	type Config,
 	type ConflictReport,
 	createBranch as createBranchFromPolicy,
+	describeNativeFinding,
 	type FunctionBundler,
+	findUndeclaredNativePackages,
 	inspect,
 	isPartialBranchCreateError,
 	loadConfigFromFile,
@@ -14,6 +17,7 @@ import {
 	PushConflictError,
 	type PushResult,
 	plan,
+	traceNativePackages,
 } from "@neon/config-runtime";
 import chalk from "chalk";
 import type yargs from "yargs";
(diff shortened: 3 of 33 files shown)


monsterzz:
@andrelandgraf for review when you have a moment. Depends on the `nativePackages` schema PR (#387) — that one should merge first.

andrelandgraf:
Andre's agent: closing this in favour of #399, for the same CI reason as #387 — a fork
`pull_request` gets no OIDC token, so `prepare` cannot reach the JFrog mirror and no job
installs. Details and evidence are on #387.

Your install / trace / verify pipeline is unchanged in substance in #399, and you are
co-author on the commit. Three things in it are corrections rather than redesign, and you may
disagree with any of them:

1. Version pinning never worked for `sharp`. `require.resolve("sharp/package.json")` throws
   `ERR_PACKAGE_PATH_NOT_EXPORTED` because its `exports` map lists only `.`, so the deploy was
   staging the registry's latest rather than the user's version. Now read off disk, with a
   regression test using that exact manifest shape.
2. The `neon function deploy --src` remediation in two error messages is not true — that path
   zips only the esbuild output and cannot carry extra files.
3. Sibling variant exclusion is decided from each package's own manifest rather than its name.
   A musl build declares `libc: ["musl"]`; a package called `musl-helper` does not, and was
   losing its files. `@img/sharp-wasm32` declares no platform metadata at all, so it stays a
   name match.

The live install tests no longer skip: CI's `prepare` points npm at the mirror, so a spawned
`npm` reaches it. They install a real `sharp` for linux-arm64 and assert the AArch64 ELF.

PR #389, opened 2026-08-05
fix(cli): serve WebSockets under neon dev

monsterzz:
## PR Checklist

- [ ] Addresses an existing open issue: fixes #000

## Overview

**WebSockets do not work under `neon dev` today, and they fail in the worst possible way — they look like they work.**

The dev server registers no `'upgrade'` listener (`packages/cli/src/dev/runtime.ts`, at the `createServer(...)` call). Node emits `'upgrade'` rather than `'request'` for anything carrying `Connection: Upgrade`, and with no listener registered it falls back to the ordinary request handler — which sees the handshake headers, does not recognise them, and answers `200 OK` on a connection the client is waiting to see a `101` on.

Verified on Node 24:

```
NO upgrade listener:   reqHandler=true  upgradeEvent=false  firstLine="HTTP/1.1 200 OK"
WITH upgrade listener: reqHandler=false upgradeEvent=true   firstLine="HTTP/1.1 101 Switching Protocols"
```

So a function's `upgrade` export was **never called** locally — `resolveFetchHandler` did not know the export existed — and the client failed against a nonsense 200. This is a bug in an already-launched contract, which is why it is worth fixing on its own.

### What changes

The dev server now handles upgrades the way the deployed runtime does, **including the precedence rule**, so local behaviour cannot disagree with deployed behaviour:

- **`export function upgrade(req, socket, head)`** is resolved by a new `resolveUpgradeHandler` (named export first, then an `upgrade` method on the default export — the same order the runtime uses) and handed the raw Node triple. This makes the launched contract work locally for the first time.
- **`upgradeWebSocket()`** from `@neon/functions` works too, reached only when there is no `upgrade` export — the legacy shape wins, exactly as deployed.
- A function that does neither gets a clean **501**, and a non-WebSocket upgrade (h2c, say) gets a **426**, rather than hanging or being answered with a bogus 200.

### Two details worth a reviewer's eye

**The upgrade path takes the raw fetch handler, not the error-boundary-wrapped one.** `withErrorBoundary` turns a throw into a 500 `Response`, which on this path is indistinguishable from a handler that deliberately declined the upgrade — so a crashing handler would have answered 501 ("no WebSocket support") and hidden the real error. The upgrade listener has its own equivalent boundary and reports a throw as the **502** the deployed runtime returns, so a user error still cannot kill the dev child process.

**The protocol implementation is duplicated from the deployed runtime, deliberately.** The runtime is a zero-dependency file baked into a microVM image and cannot import from npm, so sharing a module would mean new vendoring machinery for little gain. What the two sides actually share is the bridge global and the observable behaviour, and both are pinned by an equivalent test matrix on each side. `permessage-deflate` is not negotiated on either side.

### Tests

21 cases drive **real handshakes over loopback** with a client that speaks enough of RFC 6455 to verify masking, fragmentation, ping/pong and the close handshake:

- echo of text, binary, fragmented, and over-125-byte messages
- close from both directions (codes, reasons, `readyState` transitions, `wasClean`)
- subprotocol negotiation: echoed once when offered, header absent when not selected, throws when not offered
- the cloned-response failure — asserts a loud 500, never a silent 200
- an unmasked client frame fails the connection with 1002
- precedence when a module exports both shapes
- the Request-rebuilding path adapters take (Hono's Node adapter rebuilds it, so object identity is lost)
- the WHATWG surface: constants, `binaryType`, close-code and state guards

Plus 6 for `resolveUpgradeHandler`, including `this`-binding for the default-export method shape.

```
Test Files  2 passed (2)
     Tests  42 passed (42)
```

Typecheck and `biome check --error-on-warnings` are clean on the touched files. Three pre-existing `src/dev` test files fail to resolve unbuilt workspace deps (`@neon/config`, `@neon/config-runtime`); they fail identically on `main` and are untouched.

### Cross-repo note

This is self-contained: it needs nothing from the runtime side and is worth landing on its own, since it fixes the launched `{ fetch, upgrade }` contract locally today. The typed `upgradeWebSocket` export for `@neon/functions` is a separate PR, and the runtime that provides the upgrade when deployed lands separately again. A changeset for `neon` is included — note that a merged changeset is a version bump, not a published package.

### DIFF
--- a/.changeset/tidy-mugs-wander.md
+++ b/.changeset/tidy-mugs-wander.md
@@ -1,0 +1,13 @@
+---
+"neon": minor
+---
+
+Fix WebSockets under `neon dev`. The dev server registered no `'upgrade'` listener, so
+Node handed a WebSocket handshake to the ordinary request handler and answered `200 OK` on
+a connection the client expected to be a `101` — a function's `upgrade` export was never
+called, and the failure looked like success.
+
+The dev server now handles upgrades the way the deployed runtime does: the existing
+`export function upgrade(req, socket, head)` shape works locally for the first time, and
+`upgradeWebSocket()` from `@neon/functions` works too, with the same precedence (a legacy
+`upgrade` export wins) and the same clean `501` for a function that serves no WebSockets.
--- a/packages/cli/src/dev/runtime.test.ts
+++ b/packages/cli/src/dev/runtime.test.ts
@@ -7,6 +7,7 @@ import {
 	type FetchHandler,
 	portSelectionFromEnv,
 	resolveFetchHandler,
+	resolveUpgradeHandler,
 	withErrorBoundary,
 } from "./runtime.js";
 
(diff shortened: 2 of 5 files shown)


monsterzz:
@andrelandgraf could you take a look when you have a moment? Requesting you directly rather than via the reviewer field, which a fork account cannot set.

This is the standalone bug fix in the pair: WebSockets never worked under `neon dev` because no `'upgrade'` listener was registered, so a handshake was answered with a 200 and the function's `upgrade` export never ran. It fixes the already-launched contract locally and needs nothing from the other changes, so it can land on its own. The typed `upgradeWebSocket` export is the PR titled "feat(functions): add upgradeWebSocket for serving WebSockets from fetch".

PR #390, opened 2026-08-05
feat(functions): add upgradeWebSocket for serving WebSockets from fetch

monsterzz:
## PR Checklist

- [ ] Addresses an existing open issue: fixes #000

## Overview

Serving a WebSocket from a Neon Function today means importing `ws`, bundling it, hand-crafting the `101`, and exporting an extra `upgrade(req, socket, head)` next to `fetch` that receives raw Node objects rather than a `Request`. Nothing else in the platform works that way, and the split entrypoint breaks the single-handler mental model.

This adds the web-standard alternative, following [`Deno.upgradeWebSocket`](https://docs.deno.com/api/deno/~/Deno.upgradeWebSocket):

```ts
import { upgradeWebSocket } from "@neon/functions";

export default {
	async fetch(req: Request): Promise<Response> {
		if (req.headers.get("upgrade")?.toLowerCase() !== "websocket") {
			return new Response("expected a websocket upgrade", { status: 426 });
		}

		const { socket, response } = upgradeWebSocket(req);
		socket.addEventListener("message", (event) => socket.send(event.data));
		return response;
	},
};
```

`socket` is a standard `WebSocket`, still `CONNECTING` on return; returning `response` completes the upgrade and opens it. No extra export, no dependency, no protocol code in the bundle.

### Design

**The package stays a thin typed facade**, exactly like `waitUntil`: it reads a bridge the runtime publishes (under `Symbol.for("neon.websocket.bridge")`) and forwards to it. Every piece of protocol work — handshake, framing, fragmentation, ping/pong, the close handshake, subprotocol negotiation — lives in the runtime, so nothing here needs to change as that evolves. That is why the diff is small and the tests are about delegation rather than protocol behaviour.

**Unlike `waitUntil`, this throws off-platform rather than degrading to a no-op.** `waitUntil`'s no-op is meaningful: the promise still runs, it just isn't tracked. There is no equivalent for a WebSocket — a socket that can never open is not a degraded WebSocket, it is a bug — so an error naming the problem is the better failure.

### Two details worth knowing (both documented in the README)

**Return `response` unchanged.** A `101` cannot be expressed as a constructed `Response`: the fetch spec restricts those to 200–599 and Node enforces it, so the runtime returns an object that reports `101` and carries the pending upgrade. `clone()` drops both the status and the internal brand; rebuilding it (`new Response(res.body, res)`, as response-rewriting middleware does) throws outright. The runtime detects the lost-upgrade case and fails loudly with a named, searchable error rather than leaving the client waiting on a connection nobody upgraded.

**`binaryType` defaults to `"arraybuffer"`**, not the browser default of `"blob"` — matching Deno and other server runtimes, and what server code actually wants. `"blob"` is still accepted. `extensions` is always `""`; no extensions, including `permessage-deflate`, are negotiated.

### Backward compatibility

The existing `export function upgrade(req, socket, head)` shape is untouched and keeps working. On the runtime side it takes precedence unconditionally, so adding this API to a shared handler cannot silently re-route deployed WebSocket traffic onto a new code path.

### Tests

9 cases with a mock bridge: the happy path, that the request and options are forwarded unchanged, that the bridge is read at call time (so a later-published runtime is picked up), the off-platform throw, and that runtime-side errors propagate untouched — double upgrade, an un-offered subprotocol, and a request that is not a handshake.

```
Test Files  2 passed (2)
     Tests  14 passed (14)
```

`tsc --noEmit` and `biome check --error-on-warnings` are both clean.

### Cross-repo note

This needs a runtime that provides the upgrade. On an older runtime the export throws a clear "only available inside a Neon Functions invocation" `TypeError` rather than misbehaving, so it can ship in either order — but it is **not worth announcing until the runtime side is deployed**, since a published export that throws everywhere is worse than no export. `neon dev` support is a separate PR. A changeset is included; note that a merged changeset is a version bump, not a published package.

### DIFF
--- a/.changeset/brown-pandas-listen.md
+++ b/.changeset/brown-pandas-listen.md
@@ -1,0 +1,13 @@
+---
+"@neon/functions": minor
+---
+
+Add `upgradeWebSocket(request, options?)` for serving WebSockets from a `fetch` handler.
+
+Returns `{ socket, response }`, mirroring `Deno.upgradeWebSocket`: `socket` is a standard
+`WebSocket`, and returning `response` from your handler completes the upgrade and opens it.
+No extra export and no `ws` dependency needed — the existing
+`export function upgrade(req, socket, head)` shape keeps working unchanged.
+
+Requires a runtime that provides the upgrade (deployed, or locally under `neon dev`); on an
+older runtime it throws a clear `TypeError` rather than misbehaving.
--- a/packages/functions/README.md
+++ b/packages/functions/README.md
@@ -1,6 +1,9 @@
 # @neon/functions
 
-Runtime helpers for [Neon Functions](https://neon.com). Currently provides a `waitUntil` primitive for deferring background work past a response.
+Runtime helpers for [Neon Functions](https://neon.com):
+
+- **`waitUntil`** — defer background work past a response.
+- **`upgradeWebSocket`** — serve WebSockets from a `fetch` handler.
 
 ## Install
 
@@ -10,7 +13,7 @@ npm install @neon/functions
 
 > **Requirements:** Node.js >= 20.19.
 
-## Usage
+## `waitUntil`
 
 The API mirrors [`@vercel/functions`](https://vercel.com/docs/functions/functions-api-reference/vercel-functions-package): import `waitUntil` and call it directly with the promise you want to keep alive.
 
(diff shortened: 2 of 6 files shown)


monsterzz:
@andrelandgraf could you take a look when you have a moment? Requesting you directly rather than via the reviewer field, which a fork account cannot set.

This one is the typed export only — a thin facade over the runtime bridge, no protocol code. The companion `neon dev` fix is in the PR titled "fix(cli): serve WebSockets under neon dev", and the runtime that provides the upgrade when deployed lands separately in the dataplane repo. None of the three has to land with the others; the API degrades to a clear error rather than misbehaving in either direction.

PR #402, opened 2026-08-07
fix(cli): stop reporting an empty or borrowed identity in telemetry

nkotlyarov:
## Problem

CLI telemetry attributes runs to accounts that did not perform them, so the events cannot be
grouped by account and trusted. Nothing user-facing changes; the affected readers are whoever
queries this data.

`analyticsMiddleware` resolves an internal `userId` from whatever authenticated the invocation.
Three states got it wrong, all the same defect — the id and the key to ask the API about were
chosen by separate expressions, neither of which asked whether `ensureAuth` had selected a
credential:

- **Nothing identified the run** (no key, no credentials file recording a `user_id` — normal in CI
  and on a fresh machine). `identify` fell back with `?? "anonymous"`, and `""` is not nullish, so
  it sent `userId: ""`. `cli_command_success` sent `accountId: ""` under `authMethod: "oauth"`.
- **A key was selected explicitly.** `--api-key` and `NEON_API_KEY` record no credentials file, so
  telemetry read `DEFAULT`'s and identified the run as whoever was signed in locally, while
  `accountId` named the key's account. The borrowed id then suppressed the `getCurrentUserInfo`
  call that would have named the right one.
- **A middleware-skipped command had a key in the environment.** `profile`, `config init`, `auth`
  and a bare `neon` read `DEFAULT`'s id *and* queried the ambient key, putting two accounts in one
  event and making an API request purely to build it.

## The change

One pure function decides which credential telemetry may describe the run with. A recorded auth
context is the signal that `ensureAuth` selected something:

```ts
export const telemetryCredential = (
  authContext: AuthContext | null,
  apiKey: string | undefined,
  defaultCredentialsPath: string,
): TelemetryCredential => {
  if (authContext === null) return { credentialsPath: defaultCredentialsPath };
  if (authContext.source === "api-key") return { apiKey };
  return { apiKey, credentialsPath: authContext.credentialsPath ?? defaultCredentialsPath };
};
```

Its result is the only key queried and the only file read. Two more pure functions hold the
identity policy, replacing `userId || "anonymous"` repeated at four call sites:

```ts
export const analyticsUserId = (userId: string | undefined): string =>
  userId || ANONYMOUS;

export const storedCredentialAttribution = (
  storedUserId: string | undefined,
): EventAttribution =>
  storedUserId ? { accountId: storedUserId, authMethod: OAUTH } : {};
```

`userId` is the identity on `identify`, `CLI Started`, `CLI Error` and every `trackEvent`;
`accountId` and `authMethod` are `cli_command_success` properties only, and come from
`getAuthDetails` wherever a key was queried.

| Credential the auth middleware selected | `userId` | `accountId` / `authMethod` |
|---|---|---|
| A user-scoped key from `--api-key` or `NEON_API_KEY` | the key's own user | from `getAuthDetails` |
| An org-scoped key from `--api-key` or `NEON_API_KEY` | `anonymous` (`api_key_org` skips the lookup) | from `getAuthDetails` |
| A profile credential whose file records a `user_id` | that id | from `getAuthDetails` |
| A profile credential with no `user_id`, key user-scoped | from `getCurrentUserInfo` | from `getAuthDetails` |
| A profile credential with no `user_id`, key org-scoped | `anonymous` | from `getAuthDetails` |
| None, and `DEFAULT` records a string `user_id` | that id | that id / `oauth` |
| None, and `DEFAULT` absent, unreadable, or without one | `anonymous` | unset / unset |

The last two rows are `profile`, `config init` and `auth`, plus `dev`, `bootstrap` and `init` when
no usable credential is available. The fallback there is the `DEFAULT` file rather than the
selected profile's, so a `--profile` run of those three whose profile holds nothing usable is
attributed to whatever `DEFAULT` names — unchanged from `main`, and the documented guess.

`--help` and the single-verb help fallback exit before analytics phase 2 and are unaffected. A
bare `neon` reaches phase 2 but exits before `closeAnalytics()`, so its events are never flushed.

## Verification

- `pnpm lint:ci` clean; `pnpm test` in `packages/cli` 3874 passed, 6 skipped (same 6 as `main`).
- 11 new tests, all confirmed red before green: `storedCredentialAttribution("")` returned
  `{ accountId: "", authMethod: "oauth" }` where `{}` is expected, `analyticsUserId("")` returned
  `""`, and `telemetryCredential(null, key, …)` returned the unselected key with the default path.
- Built CLI, malformed `credentials.json` planted in a temp config dir so a read is observable:

```
profile list, ambient NEON_API_KEY   reads=1  apiCalls=0   (was apiCalls=1)
projects list --api-key <bogus>      reads=0  apiCalls=2   (was reads=1)
projects list, ambient NEON_API_KEY  reads=0  apiCalls=2
bare neon, ambient NEON_API_KEY      reads=1  apiCalls=0
--help, ambient NEON_API_KEY         middleware never runs
```

- `profile create --api-key <bogus>` makes one `/api/v2/auth` request, the handler's own
  `verifyKey`, none from telemetry.

**Not verified: the outbound payload.** `WRITE_KEY` and `host` are constants in the `Analytics`
constructor, so asserting an event on the wire needs a mock. Each half is verified separately —
the credential decision by the tests and the counts above, and Segment's serialization of an empty
`userId` by a capture against a local sink using this repo's pinned `@segment/analytics-node@1.3.0`.

## For your attention

- **No changeset**, per review: this is internal telemetry, so it ships with the next release cut.
- **Queries on this data change.** `identify.userId` never takes `''` again — that population goes
  to zero and folds into `anonymous`, which already receives one identify per run from client init.
  `accountId` becomes absent rather than empty (set to `undefined`, dropped by `JSON.stringify`), so
  `WHERE accountId = ''` stops matching rather than matching fewer rows. Key-authenticated runs move
  to the key owner's id. Nothing is backfilled.
- **Commands that authenticate inside their own handlers keep the local-default attribution.**
  `profile create --api-key`, `rotate-key` and `remove` are not authorized by the global middleware.
  Attributing them to the key their handler uses is what produced the mixed events this removes.
  Changing it means having those handlers set an auth context, which also drives 401 recovery.
- **`authMethod: "oauth"` is still a guess** on the second-to-last row, since `user_id` is stored
  for both credential kinds. It mislabels the method, not the account.
- **Telemetry payloads are untestable by construction** while the Segment host is a constant.
  Making it injectable would have caught all three defects.


.changeset/anonymous-cli-telemetry-identity.md
vadim2404, line 1: not sure that this is needed

packages/cli/src/analytics.ts
```diff
@@ -133,7 +140,10 @@ export const analyticsMiddleware = async (args: {
 				const resp = await apiClient?.getCurrentUserInfo?.();
 				userId = resp?.data?.id;
 			}
-		} else {
+		} else if (userId) {
```
vadim2404: please, add tests for these changes as well.

vadim2404 (changes requested):

packages/cli/src/analytics.ts
andrelandgraf: Andre's agent: Added — 11 tests in `packages/cli/src/analytics.test.ts`, covering the three pure functions the decision now lives in.

The branch also moved on since this comment, so the `} else if (userId)` you commented on is gone. The behavior it was reaching for is now `storedCredentialAttribution`, and two further defects of the same kind turned up and are fixed here too: a key from `--api-key`/`NEON_API_KEY` was identified as whoever was signed in locally, and a command the auth middleware skips queried an ambient `NEON_API_KEY` it never used, making an API request purely to build a telemetry event.

Tests, by behavior:

- a key no recorded authentication selected is not queried, and the local default is read instead
- a command that selected nothing falls back to the local default
- a selected key is asked about, and no file is read for it
- the file a profile selected is read, not the default one
- a stored session that recorded no path falls back to the default file
- a run nothing identified is `anonymous`, not an empty string
- an id the API never returned is treated the same as an empty one
- an identified run keeps its user id
- no account is claimed when nothing identified the invocation
- no auth method is named for an account that could not be named
- a stored session reports the account it belongs to

All 11 were confirmed failing against the old expressions before passing: `storedCredentialAttribution("")` returned `{ accountId: "", authMethod: "oauth" }` where `{}` is expected, `analyticsUserId("")` returned `""`, and `telemetryCredential(null, key, …)` returned the unselected key alongside the default path.

What is not unit-tested is the outbound payload: `WRITE_KEY` and `host: "https://track.neon.tech"` are constants in the `Analytics` constructor, so asserting an event on the wire needs a mock. Verified against the built CLI instead, counting credential-file reads and telemetry API calls with a malformed `credentials.json` planted in a temp config dir:

```
profile list, ambient NEON_API_KEY   reads=1  apiCalls=0   (was apiCalls=1)
projects list --api-key <bogus>      reads=0  apiCalls=2   (was reads=1)
projects list, ambient NEON_API_KEY  reads=0  apiCalls=2
bare neon, ambient NEON_API_KEY      reads=1  apiCalls=0
--help, ambient NEON_API_KEY         middleware never runs
```

Making the Segment host injectable would let the payload be asserted end to end, and would have caught all three defects. Not done here.

.changeset/anonymous-cli-telemetry-identity.md
andrelandgraf: Andre's agent: Your call, and it comes down to whether the behavior change is user-facing enough to announce.

The facts, so the decision is on record rather than on my reading:

- Comparable `fix(cli)` PRs in this repo ship with one — #389 and #395 both added a changeset.
- Dropping it does not hold the fix back. The code rides out with whatever release a later changeset triggers. What it loses is a version bump of its own and a changelog line.
- I claimed in the PR description that without a changeset this "never reaches a release". That was wrong and is corrected.

The case for keeping it is that two of the three fixes are observable from outside the process, not just internal bookkeeping:

- A run authenticated with `--api-key` or `NEON_API_KEY` was reported under the identity of whoever was signed in locally, so activity for one account was attributed to another.
- `neon profile …`, `neon config init`, `neon auth` and a bare `neon` sent an ambient `NEON_API_KEY` to `/api/v2/auth` purely to label a telemetry event. For a bare `neon` that request was made and the events then discarded, because `showHelp` exits before the flush. That request is gone.

A user who reads the changelog to know when their key stops being sent somewhere would want the second one. If you read that as still internal, say so and I will drop the file — it is one deletion and nothing else in the PR depends on it.

The third fix (an empty `accountId` reported under `authMethod: "oauth"`) is genuinely internal and I would not write a changelog entry for it alone.

andrelandgraf: Andre's agent: Dropped it — you were right. Internal telemetry attribution is not a changelog entry, and the file is gone in ba2dd85. It ships with the next release cut.

PR #406, opened 2026-08-08
feat(cli): inspect every database when unnamed; scope locks to the named one

andrelandgraf:
## Problem

`neon inspect db locks` and `long-running-queries` connect to one database, then read `pg_locks` and `pg_stat_activity`. Those views list sessions on the whole compute. Relation OIDs are local to the connected database, so a lock held in another database prints a null or wrong `relname`.

Omitting `--database-name` also refused to run when the branch had more than one database:

```
Multiple databases found for the branch, please provide one with the --database-name option: neondb, analytics
```

`neon diff` already treats a missing `--database` as every database on the branch. `inspect db` did not.

## Diagnosis

The session views span the compute. `pg_class` does not. That is why foreign relation locks mis-resolve.

Filtering on `pg_locks.database` would drop `transactionid` and `virtualxid` locks, which have no database. The filter that keeps those rows is the holding session: `a.datname = current_database()` on locks, `datname = current_database()` on long-running queries. Connecting once per database then makes `pg_class` resolve.

`resolveConnectionUri` still requires one database. That does not block omit-means-all. Inspect lists the branch databases from the Neon API, resolves the first connection, and rewrites the URI path for the rest. Point-in-time options stay on the query string.

Omit-means-all is the same split as `neon diff`: `--database-name` (or `neon diff --database`) limits the run to one database; omitting the flag walks every database the API lists for the branch. Database-scoped checks run once per name. Compute-wide checks (`lfc-hit-rate`, `working-set`, `replication-slots`) return the same rows from every database, so they still run once against the first listed name and keep their previous columns.

On omit, a database-scoped check always adds a `database` column, including when the branch has only one database, so adding a second database does not change the output schema.

### Help

`neon --help` now lists inspect as a branch command, not a database command:

```
neon inspect
└────────────────>    Inspect a branch's Postgres health and configuration [alia
                      ses: inspection]
```

That line was `Inspect a database's health and configuration`.

`neon inspect --help` and `neon inspect db --help` describe the work as against a branch:

```
neon inspect db
└────────────────>    Run a diagnostic query against a branch's Postgres
```

`neon inspect db --help` marks compute-wide subcommands and documents omit:

```
neon inspect db lfc-hit-rate
└────────────────>    Local File Cache hit rate (compute-wide, needs neon extens
                      ion)
neon inspect db working-set
└────────────────>    Estimated working set vs LFC size (compute-wide, needs neo
                      n extension)
neon inspect db replication-slots
└────────────────>    Replication slots (compute-wide): kind, status, client, re
                      start/confirmed-flush LSNs, and lag (pg_replication_slots
                      + pg_stat_replication)

--database-name
└────────────────>  Database to inspect. Omit to cover every database the API li
                    sts for the branch. Ranking and row limits stay per database
                    . One failing database fails the whole run. Compute-wide che
                    cks run once against the first listed database. Ignored with
                     --db-url. [string]
```

### Default table

```
neon inspect db table-sizes
```

```
┌───────────┬────────┬────────┬───────┐
│ Database  │ Schema │ Name   │ Size  │
├───────────┼────────┼────────┼───────┤
│ analytics │ public │ events │ 24 kB │
├───────────┼────────┼────────┼───────┤
│ neondb    │ public │ users  │ 16 kB │
└───────────┴────────┴────────┴───────┘
```

json and yaml use the same `database` key. Database-scoped omit sorts names with `en` `localeCompare`. `outliers` and `calls` keep `LIMIT 25` per database, not across the branch.

### Named database

```
neon inspect db table-sizes --database-name neondb
```

```
┌────────┬───────┬───────┐
│ Schema │ Name  │ Size  │
├────────┼───────┼───────┤
│ public │ users │ 16 kB │
└────────┴───────┴───────┘
```

Same columns as before. `--db-url` is this shape too.

### Locks and long-running queries

```
neon inspect db locks --database-name neondb
neon inspect db long-running-queries --database-name neondb
```

Rows are only sessions whose `datname` is that database. Empty table copy:

```
No locks held in this database.
No long-running queries in this database.
```

Omit adds `database` and uses the all-database empty copy:

```
neon inspect db locks
```

```
No locks held in any database.
```

```
neon inspect db long-running-queries
```

```
No long-running queries in any database.
```

### Errors

`--db-url` and `--database-name` leave the connect error unwrapped. Observed against `127.0.0.1:1`:

```
Could not connect to Postgres at 127.0.0.1:1: connect ECONNREFUSED 127.0.0.1:1
```

A database-scoped omit that fails to connect, on a branch with more than one database:

```
Could not connect to Postgres at 127.0.0.1:1: connect ECONNREFUSED 127.0.0.1:1 (database analytics). Pass --database-name to inspect one database.
```

A missing `neon` extension on omit:

```
This query needs the "neon" extension, which is not installed. Enable it with: CREATE EXTENSION neon; (database analytics). Pass --database-name to try a database that already has the "neon" extension.
```

A compute-wide omit that fails to connect:

```
Could not connect to Postgres at 127.0.0.1:1: connect ECONNREFUSED 127.0.0.1:1 (database analytics). Pass --database-name to connect through a different database.
```

An empty flag:

```
--database-name cannot be empty. Omit the flag to cover every database.
```

### Unchanged

- `--db-url` inspects the URL's database, never adds `database`, never suffixes the error.
- Compute-wide checks keep their previous columns.
- `outliers` and `calls` already filtered to `current_database()`. They now also fan out when the flag is omitted.

## Also in here

- `.changeset/quiet-moons-inspect.md` is `neon` minor. The default `database` column is the user-visible break.
- `neon connection-string` and inspect's URI rewrite encode the database path with `encodeURIComponent`, so a name containing `%` is not decoded on rewrite (`sales%2026` becomes `sales%252026`).
- Empty table copy for `locks` and `long-running-queries` now names the scope (`this database` / `any database`). `subscriptions` gained `No subscriptions found on any database.`

## Verification

- `vitest run src/utils/inspect_db.test.ts src/commands/inspect.test.ts`: 31 passed (target selection, error wrapping, percent encoding, point-in-time rewrite, `--help` copy, `--db-url` connect errors).
- `neon inspect --help`, `neon inspect db --help`, and `neon --help`: copy above.
- `neon inspect db locks --db-url postgresql://user:pass@127.0.0.1:1/postgres`: connect error above.

The live e2e test in the diff was not run. It creates a second database, holds locks in both, and checks named vs omit columns, lock scoping, resolvable `relname`, and a `transactionid` lock from the inspected session.

## For your attention

- The default `database` column is a compatibility break. `neon inspect db table-sizes` (and every other database-scoped check) now includes `database` in table, json, and yaml, even on a one-database branch. Parsers that assume the old field set will break. `--database-name` keeps the previous columns. The changeset is `minor`.
- Named `locks` and `long-running-queries` now return fewer rows: only the named database. That is the bugfix.
- One failing database fails the whole run. There is no partial table.
- `--db-url` does not fan out.


sav-maya:
Looks good. I think the main tradeoff would be that users investigating an unknown database must run them once per database. So it's less convenient troubleshooting if they aren't sure what database is causing issues. But since that behavior matches the existing inspect db --database-name interface, i think its pretty reasonable.

andrelandgraf:
Andre's agent: @sav-maya omit `--database-name` now covers every database the Neon API lists for the branch. `neon inspect db locks` (and the other database-scoped checks) run once per listed database and add a `database` column. Pass `--database-name` to keep the previous single-database schema.

PR #422, opened 2026-08-11
Add generated Neon API agent tools

andrelandgraf:
## Problem

The Neon Management API currently has 168 OpenAPI operations, but agent integrations have to hand-author runtime request schemas, tool metadata, and framework-specific execution wrappers. Those copies can drift from the API, silently dropped arguments can turn an agent typo into a plausible success, and HTTP values such as `Blob`, `File`, and empty responses are not valid JSON tool payloads.

The generated tools also documented `NEON_API_KEY` as required. The Management API already authenticates API keys and OAuth access tokens as the same `Authorization: Bearer` credential. Constructor-only `apiKey` made MCP session tokens unusable: `registerNeonTools` ignored `authInfo.token` and bound one client for the life of the process.

A host that already has hand-written Management API tools cannot swap them for generated ones without losing three things those tools carry: better descriptions than the OpenAPI text, a wrap around every call for tracking/spans, and grant-scoped `projectId`/`branchId` that is stripped from the published schema and filled at execute time.

## Solution

Add `@neon/tools`, generated from the same vendored OpenAPI document as `@neon/sdk`:

- strict Zod 4 request schemas for every operation, including safe OpenAPI integer bounds
- omitted fields remain omitted so the API, not a generated snapshot, applies defaults
- canonical tool descriptors backed by `@neon/sdk/raw`
- explicit operation selection with exact request and JSON-safe result types preserved through adapters
- MCP 2.x, MCP 1.x, Eve, and Mastra adapters
- JSON-safe binary, bigint, date, and empty-response handling
- generated stability, deprecation, and safety metadata

`apiKey` is a Bearer credential: a Neon API key or a Neon OAuth access token, including a getter for refresh. It is optional at construction and required at execute. MCP adapters send `authInfo.token` (MCP 2.x `http.authInfo.token`, MCP 1.x `authInfo.token`) as that credential. A present `authInfo` with an empty token is an error, not a fall back to a constructor key.

Optional host add-ons, all off by default:

- `descriptions` — replace or alter generated tool text
- `onExecute` — wrap the call for tracking or spans. `event.execute()` performs injection, original schema validation, auth, and the API request
- `inject` — fill `path.project_id` / `path.branch_id`. Without `omitFromSchema`, those keys become optional and a caller value wins. With `omitFromSchema: true`, they are removed from the published schema and the injector is the only source

The SDK spec-drift workflow now regenerates both packages, including the public request-schema exports, so an API change cannot update the SDK while leaving the tool surface stale.

## Interface

Select only the operations an agent needs:

```ts
import { createNeonTools, operationIds } from "@neon/tools";

const apiKey = process.env.NEON_API_KEY;
if (!apiKey) throw new Error("NEON_API_KEY is required");

const tools = createNeonTools({
  apiKey,
  operations: ["listProjects", "createProject"] as const,
});

const result = await tools.listProjects.execute({
  query: { limit: 10 },
});

result.data.projects; // typed Neon SDK response
```

An OAuth access token uses the same option. Pass a getter when the token can change, or pass it per call:

```ts
const tools = createNeonTools({
  apiKey: () => oauth.getAccessToken(),
  operations: ["listProjects", "createProject"] as const,
});

await tools.listProjects.execute(
  { query: { limit: 10 } },
  { apiKey: oauthAccessToken },
);
```

`operationIds` lists every valid selector. Unknown fields fail validation instead of being stripped, and invalid runtime selectors report the offending operation ID.

Use generated request schemas independently:

```ts
import { zCreateProjectBody } from "@neon/tools/schemas";

const body = zCreateProjectBody.parse({
  project: { name: "agent-project" },
});
```

The schemas are strict. If a newly added live API field is not recognized, upgrade `@neon/tools`; use `@neon/sdk` directly until a matching tools release is available.

Register the same descriptors with each framework:

```ts
import { registerNeonTools } from "@neon/tools/mcp";
import { registerNeonTools as registerNeonToolsV1 } from "@neon/tools/mcp-v1";
import { toEveTool } from "@neon/tools/eve";
import { toMastraTools } from "@neon/tools/mastra";

registerNeonTools(mcpV2Server, tools);
registerNeonToolsV1(mcpV1Server, tools);
const eveConfig = toEveTool(tools.createProject);
const mastraConfigs = toMastraTools(tools);
```

For a remote MCP server that already authenticated the client, omit `apiKey` at construction. The host must put a Neon API key or Neon OAuth access token in `authInfo.token`. This package does not implement an OAuth authorization server; that remains mcp-server-neon at mcp.neon.tech.

```ts
const tools = createNeonTools({
  operations: ["listProjects", "createProject"] as const,
});
registerNeonTools(mcpV2Server, tools);
```

Optional add-ons. Existing `createNeonTools({ apiKey, operations })` callers are unchanged:

```ts
const tools = createNeonTools({
  apiKey,
  operations: ["listProjects", "getProject", "deleteProjectBranch"] as const,
  descriptions: {
    listProjects:
      "List Neon projects in your account. Do not use for shared projects.",
  },
  onExecute: async ({ id, execute }) => {
    // host tracking / Sentry span
    return execute();
  },
  inject: {
    projectId: "project-id",
    omitFromSchema: true,
  },
});

await tools.getProject.execute({});
await tools.deleteProjectBranch.execute({ path: { branch_id: "br-id" } });
```

A getter can read request-scoped state, including a host `AsyncLocalStorage`. This package does not export one.

`mastraConfigs` retains only the selected tool IDs and their operation-specific request and result types. Eve tools use the generated snake-case ID as their filename under `agent/tools/` and default-export `defineTool(toEveTool(...))`.

Every non-read operation requires approval. Reads that return connection credentials, role passwords, or Neon Auth provider secrets are explicit approval overrides. Eve and Mastra receive their native approval fields; MCP receives standard annotations plus `neon/requiresApproval` in `_meta`.

Binary request fields accept base64 strings. Binary results remain typed and JSON-safe:

```ts
{
  data: {
    base64: "aGVsbG8=",
    contentType: "application/octet-stream",
    size: 5,
  },
}
```

## Also in here

- Companion publish allowlist: [databricks/secure-public-registry-releases-eng#206](https://github.com/databricks/secure-public-registry-releases-eng/pull/206) adds `@neon/tools` to the neon-pkgs.yml dispatch choices. This PR does not publish the package.
- The package starts at `0.0.0` with a minor changeset for its first `0.1.0` release, plus a second minor changeset for the host add-ons.
- CI runs the Eve compatibility suite on Node.js 24.
- Distribution type checks now cover every `@neon/tools` export.
- The package supports Node.js 20.19; Eve requires Node.js 24 and Mastra requires Node.js 22.13.

## Verification

- `pnpm --filter @neon/tools tsc && pnpm --filter @neon/tools test:ci && pnpm --filter @neon/tools build`
- 87 package tests covering all 168 operation IDs, strict generated JSON Schemas, omitted defaults, merged OpenAPI object shapes, typed selection and adapter conversion, typed JSON-safe results, OpenAPI integer bounds, GET/POST requests, approval metadata, binary input/output, empty responses, typed API errors, cancellation, MCP 1.x/2.x discovery and calls, Bearer API-key and OAuth tokens, credential getters, execute-time overrides, concurrent credential isolation, missing-credential failures, MCP `authInfo` forwarding that fails closed on an empty token, description overrides on `createNeonTools` and `createNeonTool`, `onExecute` wrapping including hook-thrown and API failures, grant-locked injection that ignores mutated `event.input`, fill vs omit published schemas, mixed project/branch injection, query/body preserved beside injected path, `getConnectionURI` query `branch_id` left alone, host `AsyncLocalStorage` getter isolation, MCP v1/v2 omitted and fill-mode calls, Eve/Mastra omitted execute, tools with no path key, empty/missing/non-string injectors, and invalid injected ids rejected before fetch
- package tests passed under Node.js 22
- `pnpm exec attw --pack packages/tools --profile esm-only` (from the earlier commits on this branch)
- live Neon smoke test on this branch created, listed, and deleted a project through generated tools with an org-scoped API key; that key was revoked
- OAuth was not exercised against a live Neon token. Unit tests send OAuth-shaped Bearer credentials and MCP `authInfo.token` through the same path the SDK uses for API keys.
- Path injection was not exercised against a live Neon project. Unit tests assert the request URL after mock fetch.

## For your attention

- npm publish is not this PR. The companion allowlist is [databricks/secure-public-registry-releases-eng#206](https://github.com/databricks/secure-public-registry-releases-eng/pull/206). After this merges at `0.0.0`, bootstrap `@neon/tools@0.0.0` on npm, configure Trusted Publisher, then cut the Changesets bump.
- MCP annotations and `_meta` are advisory. The host must enforce approval before execution.
- Non-read operations are intentionally classified conservatively as potentially destructive, so some idempotent writes will still prompt.
- Sensitive-read approval is an explicit operation-level policy. OpenAPI `x-sensitive` cannot derive it because that marker also covers ordinary PII and fields omitted by specific endpoint responses.
- `authInfo.token` is forwarded as a Neon Bearer credential. The host must populate it with a Neon API key or Neon OAuth access token. `@neon/tools` does not verify, refresh, or issue OAuth tokens.
- A present MCP `authInfo` with an empty token is an error. It does not fall back to a constructor key.
- These add-ons do not flatten `{ path, query, body }` into camelCase `projectId`, do not rename tools (`get_project` is not `describe_project`), and do not add host-only behavior such as returning a connection string from `createProject`.
- `branchId` injection fills `path.branch_id` only. Query/body branch selectors such as `getConnectionURI`'s `query.branch_id` are unchanged.
- Grant filtering, read-only filtering, and access-control notices stay in the host.
- Injection is not authorization. Project-agnostic operations and query/body project ids remain the host's responsibility.
- No mcp-server-neon migration is in this PR.


andrelandgraf:
Companion publish allowlist: https://github.com/databricks/secure-public-registry-releases-eng/pull/206

PR #425, opened 2026-08-11
Default project and branch commands to list

andrelandgraf:
## Problem

Bare resource groups currently show help, so listing requires an explicit subcommand:

```console
$ neon projects
neon projects <sub-command> [options]

Commands:
neon projects list
└────────────────>    List projects
neon projects create
└────────────────>    Create a project
...
```

A generic top-level `neon list` shortcut would be shorter, but it would not name the resource and would be ambiguous now that projects and branches are both primary CLI resources.

## Solution

Make `list` the default subcommand for the existing `projects` and `branches` resource groups. Yargs registers `$0` as an alias of `list`, so the default and explicit forms share one builder and handler without exposing the internal alias in shell completion.

The explicit `list` forms remain supported. The generic top-level `neon list` command is not added.

## Interface

```bash
# Projects: equivalent forms
neon projects
neon projects list

# Branches: equivalent forms
neon branches --project-id <project-id>
neon branches list --project-id <project-id>
```

The fixture-backed bare project command preserves both owned and shared output sections:

```console
$ neon projects --output yaml
projects:
  - id: 1
    name: Project_1
    created_at: 2019-01-01T00:00:00.000Z
    updated_at: 2019-01-01T00:00:00.000Z
  - id: 2
    name: Project_2
    created_at: 2019-01-01T00:00:00.000Z
    updated_at: 2019-01-01T00:00:00.000Z
  - id: 3
    name: Project_3
    created_at: 2019-01-01T00:00:00.000Z
    updated_at: 2019-01-01T00:00:00.000Z
shared_with_you:
  - id: adj-noun-12401747
    region_id: aws-us-east-2
    name: Shared Project
    created_at: 2024-04-03T04:45:46Z
    updated_at: 2024-04-11T16:13:43Z
```

The existing singular aliases also default to listing:

```bash
neon project
neon branch --project-id <project-id>
```

Help remains a network-free command-group action and marks the list command as the default:

```console
$ neon projects --help
Commands:
neon projects list
└────────────────>    List projects [default]
neon projects create
└────────────────>    Create a project
neon projects update <id>
└────────────────>    Update a project
...
```

`neon projects` accepts the existing list flags, including `--org-id` and `--recoverable-only`. `neon branches` resolves project context exactly as `neon branches list` does, including `--project-id` and a linked `.neon` file. Shell completion offers `list`, not the internal `$0` alias.

## Also included

- Document both default-list forms and their project-list filtering behavior in the CLI README.
- Add a minor changeset that discloses the previous bare-command help behavior.

## Verification

- `pnpm --filter neon build`
- `pnpm --filter neon lint`
- `pnpm --filter neon test` — 4,105 passed; 6 pre-existing skipped
- Default project/branch, help, and completion tests — 77 passed
- Plural and singular bare forms return list output.
- Project defaults preserve `--org-id` and `--recoverable-only`.
- Branch defaults preserve explicit `--project-id` and linked `.neon` context.
- `--help` exits successfully against an unreachable API host and still shows non-list subcommands.
- Completion includes `list` and excludes `$0` for both resource groups.
- `pnpm test:e2e:live` — 52 live tests passed across SDK (16), config (2), config-runtime (2), env (1), and CLI (31).

The live runs used temporary organization-scoped API keys. The suites removed their projects and both keys were revoked after use.

## For attention

- Bare `neon projects` and `neon branches` previously showed group help. They now require authentication and perform the list action; use `--help` for discovery. An unauthenticated interactive run can therefore start the CLI's documented login flow, while CI exits non-zero instead of printing help.
- This PR intentionally stops at projects and branches. Whether default listing should extend to project-scoped databases, roles, and other resource groups is not decided here.
- Analytics preserves the raw command names (`projects`, `project`, `branches`, or `branch`) and keeps the existing list attribution behavior.


andrelandgraf:
Closed: help is better than a list shorthand

PR #426, opened 2026-08-11
Add --project-id/--org-id/--branch-id to neon init

jeff-at-neon:
## Summary
- Adds three optional flags to `neon init` — `--project-id`, `--org-id`, `--branch-id` — that streamline setup for users who already have a Neon project, in both agent and interactive modes.
- When `--project-id` is provided, the getting-started flow collapses its org/project select-or-create-or-context chain into a single **verified** `neon link` step (verifies the IDs against the account, infers the org, records the branch, and writes `.neon`). `--org-id` alone scopes project selection to that org and skips org selection; `--branch-id` is recorded in `.neon` so `neon env pull` targets that branch.
- IDs are validated at the CLI boundary (new shared `assertSafeId` in `src/init/ids.ts`) before any shell interpolation, so a malformed ID is rejected early in the caller's expected shape (JSON in agent mode).
- The `--data '{"step":"db"}'` agent path also gains `branchId`, threaded into the connection-string command and `project_ready` payload.

### Behavior
| Provided | Effect |
|---|---|
| `--project-id` | Skips org + project selection; one verified `link`, then env pull |
| `--org-id` alone | Skips org selection; still lists/creates a project in that org |
| `--branch-id` | Recorded in `.neon` so env pull targets that branch |

The no-ID path is byte-identical to before (only the `--help` snapshot changed, to list the new flags).

## Test plan
- [x] `pnpm typecheck` (tsc --noEmit)
- [x] biome check (error-on-warnings) clean on changed files
- [x] `vitest run` init suites — 310/310 pass (`agent_snapshot`, `init`, `route_command`, `orchestrate`)
- [x] Regenerated `--help` snapshot (adds the three flags only)
- [x] Smoke test on built binary: flags appear in `--help`; malformed `--project-id` rejected as JSON in agent mode
- [ ] Consider whether a changeset is needed for the `neon` package release

### DIFF
--- a/packages/cli/src/commands/init.test.ts
+++ b/packages/cli/src/commands/init.test.ts
@@ -77,6 +77,9 @@ describe("init", () => {
 			agent: "cursor",
 			skipMigrations: undefined,
 			preview: undefined,
+			projectId: undefined,
+			orgId: undefined,
+			branchId: undefined,
 		});
 		expect(interactiveInit).not.toHaveBeenCalled();
 	});
@@ -96,6 +99,9 @@ describe("init", () => {
 			agent: "claude",
 			skipMigrations: true,
 			preview: undefined,
+			projectId: undefined,
+			orgId: undefined,
+			branchId: undefined,
 		});
 	});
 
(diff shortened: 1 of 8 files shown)


jeff-at-neon:
Superseded by #428, which makes `neon init` minimal (install tooling, then hand off to the agent). The `--project-id/--org-id/--branch-id` flags this PR added are removed there since `init` no longer links projects.

PR #428, opened 2026-08-11
Make `neon init` minimal: install tooling, then hand off to the agent

jeff-at-neon:
## Summary

Makes `neon init` minimal: it authenticates and installs the Neon tooling (MCP server, agent skills, and — in a VS Code–based IDE — the editor extension), then hands off to the agent. It no longer selects an organization or project, links the directory, pulls environment variables, asks which features to enable, or provisions Auth / Object Storage / Functions / AI Gateway. Connecting a database and configuring features is driven by the installed Neon skill (or a manual `neon link`).

## Changes

- **Orchestrator** (`orchestrate.ts`): reduced to `auth → inspect(mcp + skills) → setup(install) → hand-off`. No `.neon` reads/writes, no connection-string resolution.
- **Setup** (`phases/setup.ts`): installs the Neon CLI, MCP server, skills, and (optionally) the editor extension, then returns a terminal hand-off. Dropped feature selection, preview mode, template scaffolding, and `.neon` writes.
- **Interactive** (`interactive.ts`): installs tooling, then directs the user to finish in their agent or run `neon link`. Removed the template picker, feature prompt, and Neon Auth handling.
- **Flags removed** from `neon init`: `--preview`, `--project-id`, `--org-id`, `--branch-id`, `--skip-migrations`.
- **Agent `--data` protocol** reduced to `auth`, `setup`, `mcp`, and `skills`. Removed the `getting-started`, `db`, `migrations`, `neon-auth`, `status`, and `finalize` steps.
- **Deleted** phases/support modules: `getting_started`, `neon_auth` (the phase only — the top-level `neon neon-auth` command is unchanged), `migrations`, `db`, `cleanup`, `status`, `resolve_context`, `ids`.
- The standalone `neon bootstrap` command is unaffected.

## Tests

- Rewrote `orchestrate`, `setup`, `route_command`, and `init` unit tests for the minimal flow and pruned the `agent_snapshot` matrix (removed-step cases dropped, snapshot regenerated).
- Updated the init e2e test: the removed org/project/env sequence is replaced with a check that the `setup` step still answers with an `agent_check` and installs nothing into the project just by being asked; the unknown-step guard is kept.

This pull request and its description were written by Isaac.

PR #431, opened 2026-08-13
Improve neon init onboarding and preserve fields on .neon writes

philip:
## Update

Split this up as requested. The three independent pieces are now their own PRs off main:

- #432 install the `neon` CLI, not `neonctl`. This is the isolated one to merge now.
- #433 preserve unrelated fields on `.neon` writes. The non-destructive context write.
- #434 point init's doc links at canonical neon.com/docs pages.

Each is single-purpose and reviewable on its own. #432 and #433 have no dependencies; #434 is for doc URLs. Setting #431 to Draft.

What remains here is the fourth piece: rewiring getting-started to delegate to neon link (plus the neon psql connection check and installing deps before env pull). I stopped here as it depends on #433 for correctness and overlaps with the minimal-init redesign in #428.

## Overview

`neon init` had drifted from the current CLI. It probed for the retired
`neonctl` alias, so on machines with the `neon` CLI installed it failed to
detect it and tried to reinstall; it hand-edited `.neon` (missing the branch)
rather than using `neon link`; and it referenced doc pages that had since moved.

This reworks the standard onboarding flow to detect and install `neon`, delegate
org/project selection and the `.neon` write to `neon link --agent` (which records
org, project, and branch), point the doc references at current pages, and suggest
`neon psql` as a quick connection check alongside the driver-based verification.
Dependencies are now installed before linking, so any env pull can safely import a
project `neon.ts`. The `--preview` path is left as is.

Separately, and more broadly than init: writing `.neon` through `link`, `checkout`,
or `set-context` now preserves fields it does not own, so a single `neon link` no
longer overwrites init's in-progress state. `neon link --clear` still resets the file.

Verified end-to-end against a live backend: the rewritten init e2e test drives
`neon link` through project creation, confirms `.neon` gets org + project + branch,
and that `env pull` writes a real `DATABASE_URL`.

## Follow-up (not in this PR)

Making init fully backend-friendly is deferred: dropping the `--preview` gating so
the bootstrap path works for everyone once the platform features are GA, and adding
a plain-Postgres bootstrap template (today's starters all require Functions, Object
Storage, or the AI Gateway). That template also needs a matching example added to
`neondatabase/examples`. Relatedly, init could scaffold a `neon.ts`, and the services
it declares under `defineConfig`'s `preview` field should work without that `preview`
grouping once GA.

The `db` and `status` init phases look like leftovers: the orchestrator routes
onboarding to `getting-started` and never to them, and nothing in the repo chains to
`status`, yet they still carry the old hand-rolled project-creation pattern. Pending
a check that no external skill references the `db` step, they are candidates for
removal.

### DIFF
--- a/.changeset/neon-init-link-and-context.md
+++ b/.changeset/neon-init-link-and-context.md
@@ -1,0 +1,10 @@
+---
+"neon": minor
+---
+
+Improve `neon init` onboarding and make `.neon` context writes non-destructive:
+
+- Detect and install the `neon` CLI during init, not the retired `neonctl` alias.
+- Use `neon link` to create/link the project and write `.neon` (branch included) instead of hand-editing it.
+- Writing `.neon` via `link`, `checkout`, or `set-context` now preserves unrelated fields; `neon link --clear` still resets it.
+- Update init's doc links, and suggest `neon psql` as a quick connection check.
--- a/packages/cli/e2e/init.e2e.test.ts
+++ b/packages/cli/e2e/init.e2e.test.ts
@@ -10,12 +10,7 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { configuredOrgId } from "@neon/e2e-harness";
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
-import {
-	createProject,
-	deleteProject,
-	runCli,
-	uniqueProjectName,
-} from "./helpers.js";
+import { deleteProject, runCli, uniqueProjectName } from "./helpers.js";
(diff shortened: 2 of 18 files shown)


philip:
I created this before seeing https://github.com/neondatabase/neon-pkgs/pull/428 which changes things. Still, if 428 is merged, there are bits here worth merging (e.g., neonctl->neon and the link fixes).

philip:
The `init` process was written via https://github.com/neondatabase/neon-pkgs/pull/485

PR #433, opened 2026-08-14
Preserve unrelated fields when writing the .neon context file

philip:
## Overview

`updateContextFile` wrote `.neon` wholesale, replacing the entire file on every write. That means a single `neon link` (or `checkout`, or `set-context`) would clobber the ephemeral `_init` state `neon init` stashes in the same file, along with anything a user had added by hand. This is the blocker that stops `neon init` from delegating to `neon link` for project setup.

Fix it by merging instead of overwriting:

- The managed keys (`orgId`, `projectId`, `branch`, `branchId`) are still fully governed by the write, so re-linking without a branch still clears a stale one.
- Every other ("foreign") key already in the file is read back and carried forward.
- `neon link --clear` gets a dedicated `clearContextFile` that still resets the file wholesale, since "forget this directory" should drop everything.

`neon project create` already calls `updateContextFile` and now benefits from the same preservation without changes; its managed-key behavior is unchanged.

Split out of #431 so it can be reviewed on its own.

### DIFF
--- a/.changeset/neon-context-preserve-fields.md
+++ b/.changeset/neon-context-preserve-fields.md
@@ -1,0 +1,5 @@
+---
+"neon": patch
+---
+
+Writing the `.neon` context file via `link`, `checkout`, or `set-context` now preserves fields the command doesn't own instead of overwriting the whole file. Managed keys (`orgId`, `projectId`, `branch`, `branchId`) are still governed by the write, but foreign keys, most importantly the ephemeral `_init` state `neon init` stashes, are carried forward. `neon link --clear` still resets the file wholesale.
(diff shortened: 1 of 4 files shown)


andrelandgraf:
Why would someone write their own content to `.neon`?

philip:
> Why would someone write their own content to `.neon`?

This was originally part of PR #431 (init update), and `init` did add custom `.neon` steps during the flow, which `link` clobbered. But that's when `init` manually edited `.neon`. That was the purpose, other than that users do all sorts of things but I'll close this given the original intent. Also now I see `claim` adds a `.neon` option but it doesn't allow `link` anyways.

PR #434, opened 2026-08-14
Point neon init doc links at canonical neon.com/docs pages

philip:
## Overview

`neon init` surfaced documentation hints (the `prerequisite` and `context` fields agents follow) using the old `/docs/ai/skills/neon-postgres/references/*.md` skill files. Those pages were retired and now only redirect, so this points them at the canonical docs directly: getting started, connection methods, Neon Auth, the serverless driver, the CLI, branching, and the JS SDK.

It also renames the map and its emitted field from `skillReferences` to `docReferences`. These are documentation links, not the agent skill files installed by the skills CLI, so the old name implied a connection that does not exist. The emitted-field rename is a small change to the agent protocol output.

Verified all eight new URLs resolve (200) and the old ones 308-redirect onto live pages, so this removes a redirect hop.

Split out of #431 so the doc-link update is easy to review on its own.

### DIFF
--- a/.changeset/init-canonical-doc-urls.md
+++ b/.changeset/init-canonical-doc-urls.md
@@ -1,0 +1,5 @@
+---
+"neon": patch
+---
+
+Point `neon init`'s doc hints at canonical `neon.com/docs` pages instead of the retired skill-reference URLs, and rename the `skillReferences` field to `docReferences`.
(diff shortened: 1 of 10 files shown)


philip:
The skills code was refactored after I wrote this, closing.

PR #451, opened 2026-08-20
feat(cli): add stalled query diagnostics

sav-maya:
## Problem

When a compute is struggling, the question during an incident is what is running right now and what is holding it up.

`neon inspect db long-running-queries` answers part of that. It lists queries past 5 minutes in the current database, with pid, duration, state and query.

Three parts of the incident question aren't covered yet. Five minutes is well past the point where you want to be looking. Parallel workers arrive as separate rows carrying the same query text, with nothing tying them to the backend that spawned them. And no column says which backend is waiting on which, so the next step is opening psql and writing `pg_stat_activity` joins by hand.

## The command

```bash
neon inspect db stalled-queries
neon inspect db stalled-queries --output json
```

It takes a read-only snapshot of backends in `pg_stat_activity` that are `active`, are a `client backend` or a `parallel worker` and belong to a query group where something has been running for at least 30 seconds. The inspecting session excludes itself with `pg_backend_pid()`.

Table output on an 80-column terminal, one query running with two parallel workers:

```
Duration         Wait Event        Role    Query Group  Query
00:00:41.923425  BgWorkerShutdown  leader  952          SELECT customer_id, s...
00:00:41.923425  DataFileRead      worker  952          SELECT customer_id, s...
00:00:41.923425  DataFileRead      worker  952          SELECT customer_id, s...
```

The same snapshot with a fourth backend waiting on a lock:

```
Duration         Wait Event        Blocking Pids  Role    Query Group  Query
00:00:41.923425  BgWorkerShutdown                 leader  952          SELECT...
00:00:41.923425  DataFileRead                     worker  952          SELECT...
00:00:41.923425  DataFileRead                     worker  952          SELECT...
00:01:12.005118  transactionid     771            leader  1104         UPDATE...
```

Nothing running past the threshold:

```

No active queries running longer than 30 seconds.
```

## Reading a row

`query_group` is `COALESCE(leader_pid, pid)`, so a leader and its parallel workers share one value and sort together, leader first. `role` is `leader` when `leader_pid` is null and `worker` otherwise. A group qualifies when any backend in it passed 30 seconds, and then every member of the group is returned, including a worker that started later.

The `activity` CTE is `MATERIALIZED`, so `pg_stat_activity` is read once and the group join sees a single snapshot.

`duration` and `observed_at` come from `statement_timestamp()`, which is the moment the snapshot is taken rather than the start of the surrounding transaction.

The table columns are ordered `duration`, `wait_event`, `blocking_pids`, `role`, `query_group`, `query`. The table writer only accepts a narrow-terminal fit where the last column shrank, so putting `query` last keeps the five diagnostic columns intact and truncates the query text instead.

Blockers come from `array_to_string(pg_blocking_pids(a.pid), ',')`. That produces an empty string when nothing is blocking the backend, and the table writer drops a column whose value is empty in every row, so `Blocking Pids` appears only when something is actually blocked.

`--output json` and `--output yaml` dump the full SQL row:

```json
[
  {
    "observed_at": "2026-08-20 21:27:52.147991+00",
    "query_start": "2026-08-20 21:27:10.224566+00",
    "query_group": "1104",
    "pid": "1104",
    "leader_pid": null,
    "role": "leader",
    "backend_type": "client backend",
    "database": "neondb",
    "application_name": "billing-worker",
    "query_id": "1189275022371128103",
    "state": "active",
    "wait_event_type": "Lock",
    "wait_event": "transactionid",
    "blocking_pids": "771",
    "duration": "00:01:12.005118",
    "query": "UPDATE invoices SET status = 'paid' WHERE id = $1"
  }
]
```

A blocking pid can belong to an idle-in-transaction backend, which this command does not list, because the snapshot is filtered to `state = 'active'`. `neon inspect db locks` shows lock holders.

The query is registered with `scope: "compute"`, so it runs once against one database and the output carries no database column. `--project-id`, `--branch`, `--database-name`, `--role-name` and `--db-url` are the shared `inspect db` options and behave as they do for every other subcommand. Nothing changes for the existing subcommands.

## Also in here

- `packages/cli/README.md` gains a `Database diagnostics (inspect)` section and an `inspect` row in the command table. The README had no `inspect` section before, so this documents the new subcommand rather than the whole group.
- The changeset is `"neon": minor`.
- The `inspect db` wiring loop covers every key in `INSPECT_QUERIES`, so the new entry adds one snapshot line for the shared unreachable-`--db-url` assertion.

## Verification

`npx vitest run src/commands/inspect.test.ts` in `packages/cli`, 20 tests pass, against base `0907e46`.

Unit behaviours added:

- the entry is compute-scoped, calls `pg_blocking_pids` and filters to `client backend` and `parallel worker`, and keeps the six fields in display order
- at 80 columns the table shows Duration, Wait Event, Role, Query Group and Query, truncates only the query text and no line exceeds 80 columns
- a blocked row brings back the Blocking Pids column with the blocking pid in it, still within 80 columns
- JSON output keeps every field the SQL returns, not the six table fields
- the subcommand is reachable as `inspect db stalled-queries` and reports a Postgres connection error on an unreachable `--db-url`

Live e2e in `packages/cli/e2e/inspect.e2e.test.ts`, which creates a real project and runs under `pnpm test:e2e`:

- a fresh project returns `[]`
- a backend running `SELECT pg_sleep(90)` appears after a 32 second wait, with `role` `leader`, `query_group` equal to its own pid, `state` `active` and a string `duration`, then gets cancelled

Not covered live: the leader and worker grouping. The e2e drives a single backend, so worker rows are exercised only through the unit tests, which feed the writer rows shaped like the SQL output. Reproducing real parallel workers needs a plan the planner chooses to parallelise.

## For your attention

- **The 30 second threshold is fixed.** There is no flag for it, which keeps the subcommand consistent with the rest of `INSPECT_QUERIES`, where each entry is a fixed query. A different threshold means editing the SQL.
- **A blocker can be absent from the output.** Blocking pids are reported for every listed backend, but the blocker itself is only a row when it is active and past the group threshold. An idle-in-transaction blocker shows up as a pid with no matching row.
- **`--database-name` only picks the connection route.** The view is compute-wide, so the flag changes which database the CLI connects through and not which rows come back.
- **The live suite gets 32 seconds longer.** The threshold can only be crossed by waiting, and the test allows 240 seconds.
- **Query text is truncated in the table on narrow terminals.** The full text is in `--output json` and `--output yaml`.


andrelandgraf:
Andre's agent: One note on the default table output.

Terminal width is limited. The list writer from #446 (`packages/cli/src/human_table.ts`, contract in `packages/cli/AGENTS.md`) measures each row against the terminal and drops trailing fields until it fits. Only the last surviving column shrinks with `...`.

That makes column order the priority order. `stalled-queries` lists 16 fields and puts waits, blockers, duration and query last, so those are the ones that disappear on a normal-width terminal. Move them to the front. `long-running-queries` already leads with duration, state and query, and `profile list` moved Credentials ahead of Scope in #446 for the same reason.

Then cut what's left over. The other inspect commands sit at 2-8 fields. A field that never survives the width check is a field nobody sees in `-o table`, and `--output json` still carries every one of them, so removing a column from the table loses no data.

andrelandgraf (approved):
Andre's agent: Engineering and DX review both approved. Merging.

PR #489, opened 2026-08-26
fix: separate resource creation from credential retrieval

### LINKED ISSUE #369: [Bug]: neon projects create always prints the full connection string including the password, no quiet/silent flag
## What happened

Running `neon projects create` (and the MCP server's `create_project` tool) always returns the full Postgres connection string, password included, in both table and JSON output. There's no flag to suppress this.

## Why this matters

This makes it unsafe to run project creation from any context where output might be logged, screen-shared, or passed through an AI coding agent, since the secret is unavoidably exposed in the tool's response. I hit this directly running Claude Code against a Neon project: the only safe way to avoid the password reaching the agent's context is to have a human run the command in their own terminal, which defeats a lot of the value of CLI/MCP automation for this one step.

## What I'd like to see

- A `--quiet` / `--no-secrets` flag on `neon projects create` that omits the password from stdout (a connection string with the password redacted, or a placeholder, would still confirm success).
- The same option on the MCP `create_project` tool: return the connection string with the password redacted by default, requiring a separate explicit call to retrieve the actual secret.

Every other credential-adjacent Neon CLI command I've used doesn't have this problem since it doesn't return a plaintext secret by default; project creation specifically always does.

agustin-caceres:
## PR Checklist

- [x] Addresses an existing open issue: fixes #369

## Overview

I ran into #369 while using the CLI through an agent workflow. The agent only needed the new project ID, but project creation also handed it the full connection string. This PR gives callers a way to create the resource without pulling credentials into the same output.

On the CLI side, `neon projects create` and `neon branches create` now accept `--no-secrets`. Their existing output stays the default for backward compatibility, while commands emitted by `neon init --agent` use the secret-free form automatically.

For `@neon/tools`, creation and credential retrieval are separate operations. The published `projects.create` and `branches.create` tools return only resource metadata; callers can use the approval-gated `postgres.connectionString` tool later if they actually need credentials. This replaces the published `projects.createAndConnect` and `branches.createWithCompute` selectors, and the README examples now reflect that flow.

The regression coverage feeds credential-bearing API responses into the CLI and tool adapters, then checks that the sentinel password never reaches `--no-secrets`, direct tool, MCP v1, or MCP v2 output. It also keeps an explicit check that `postgres.connectionString` still returns the credential when requested.

## Validation

- `pnpm --filter neon... build`
- `pnpm --filter @neon/tools build`
- `pnpm --filter @neon/tools generate:check`
- `pnpm --filter @neon/tools test -- --run` (126 tests passed)
- focused `neon init` phase suites (23 tests passed)
- built CLI verified in table, JSON, and YAML output modes
- Biome checks on the changed source files and `git diff --check`

### DIFF
--- a/.changeset/safe-agent-creation.md
+++ b/.changeset/safe-agent-creation.md
@@ -1,0 +1,7 @@
+---
+"@neon/tools": minor
+"neon": minor
+"neonctl": minor
+---
+
+Publish `projects.create` and `branches.create` as credential-free creation tools, with connection strings retrieved explicitly through `postgres.connectionString`. The combined `projects.createAndConnect` and `branches.createWithCompute` selectors are no longer published. Add `--no-secrets` to project and branch creation in the Neon CLI.
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -133,6 +133,18 @@ automation:
 neon branches create --project-id <project-id> --name production --protected
 ```
 
+Project and branch creation include connection credentials in their output by
+default for backward compatibility. When the output may be logged or passed to
+an agent, use `--no-secrets` to return only the created resource metadata:
+
+```bash
+neon projects create --name my-project --output json --no-secrets
+neon branches create --project-id <project-id> --name preview --output json --no-secrets
+```
+
+The flag omits the complete `connection_uris` block rather than redacting one
+field. Retrieve a connection string separately when it is actually needed.
+
 ### Enable logical replication
 
 Enable logical replication for every endpoint in an existing project:
--- a/packages/cli/mocks/main/projects/POST.js
+++ b/packages/cli/mocks/main/projects/POST.js
@@ -1,7 +1,32 @@
 import { expect } from 'vitest';
 
 export default function (req, res) {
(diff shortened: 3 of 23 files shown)


andrelandgraf:
Andre's agent: Thanks for this. The split you named is the right one: create the resource without putting credentials in the agent transcript, then fetch a connection string only when something actually needs one.

We are landing that in two follow-ups rather than merging this PR:

- MCP: https://github.com/neondatabase/mcp-server-neon/pull/324 — `create_project` and `create_branch` will call `projects.create` / `branches.create`, so the tool result has no connection string. `get_connection_string` stays for the cases that need a URI.
- CLI: a follow-up PR that adds `--no-secrets` on `neon projects create` and `neon branches create`, and points `neon init --agent` at that form.

The SDK and tools side of the same split already landed in https://github.com/neondatabase/neon-pkgs/pull/493 (`create` vs `createAndConnect`; branch create still attaches compute by default). That differs from publishing a compute-less `create` here, so we are not taking this branch as-is.

Closing this PR in favor of those two. If `--no-secrets` as a flag (default on) is the wrong CLI tradeoff, say so on the follow-up and we will change it.

PR #550, opened 2026-09-05
Return `{ branch, endpoint?, connectionString? }` from `neon.branches.create`

andrelandgraf:
## Problem

`neon.branches.create` returned only the `Branch`. The Neon API's create-branch response already includes the endpoints it provisioned and connection URIs, but the SDK dropped both. Anyone who wanted a host or connection string after creating a branch had to make a second round trip via `endpoints.listByBranch` plus `postgres.connectionString`, or switch to `createAndConnect`, which errors when no URI comes back.

## Diagnosis

The create response mapper picked `branch` off the API payload and discarded `endpoints` and `connection_uris`. `createAndConnect` had its own mapping that read those fields, so the data was available — it just wasn't surfaced through the plain `create` path.

## Interface

Before:

```ts
const { data, error } = await neon.branches.create(projectId, { name: "preview" });
data; // Branch
// host/URI required endpoints.listByBranch + postgres.connectionString, or createAndConnect
```

After:

```ts
const { data, error } = await neon.branches.create(projectId, { name: "preview" });
data?.branch;
data?.endpoints; // all endpoints the API returned
data?.endpoint; // endpoints[0]
data?.connectionUris; // all URIs the API returned
data?.connectionString; // pooled form of connectionUris[0]; undefined if none

const connected = await neon.branches.createAndConnect(projectId, { name: "preview" });
// still { branch, endpoint, connectionString }; client error if no URI
```

- `createAndConnect` contract is unchanged.
- No `pooled?` option on `create` — `connectionString` is the pooled form of the first URI.
- `BranchCreateResult` is exported from `@neon/sdk`.

## Also in here

- `@neon/tools` `branches.create` returns the same shape.
- E2e tests updated to read `.branch.id` instead of `.id`.
- Major changeset for `@neon/sdk` and `@neon/tools`.

## Verification

- `pnpm --filter @neon/sdk test:ci` — 144 passed (includes new `create` / `createAndConnect` tests)
- `pnpm --filter @neon/sdk test:types` — 16 passed
- `pnpm --filter @neon/sdk build`
- `pnpm --filter @neon/tools exec vitest run src/workflows.test.ts` — 19 passed
- Live e2e not run; files were updated so they compile against the new shape.

## For your attention

- Breaking: callers reading `data.id` after `create` must move to `data.branch.id`.
- `connectionString` is `undefined` rather than an error when the API omits a URI. `createAndConnect` remains the strict variant.
- The `pooled?` option was left off `create`. Extra endpoints and URIs stay on `endpoints` and `connectionUris`.

### DIFF
--- a/.changeset/sdk-branch-create-result.md
+++ b/.changeset/sdk-branch-create-result.md
@@ -1,6 +1,6 @@
 ---
-"@neon/sdk": minor
-"@neon/tools": minor
+"@neon/sdk": major
+"@neon/tools": major
 ---
 
 `neon.branches.create` now returns `{ branch, endpoint?, connectionString? }` instead of a bare `Branch`. The pooled URI is included when the API returns one. `createAndConnect` still requires a URI. The `branches.create` tool returns the same shape.
--- a/.changeset/sdk-branch-create-result.md
+++ b/.changeset/sdk-branch-create-result.md
@@ -3,4 +3,4 @@
 "@neon/tools": major
 ---
 
-`neon.branches.create` now returns `{ branch, endpoint?, connectionString? }` instead of a bare `Branch`. The pooled URI is included when the API returns one. `createAndConnect` still requires a URI. The `branches.create` tool returns the same shape.
+`neon.branches.create` now returns `{ branch, endpoints?, endpoint?, connectionUris?, connectionString? }` instead of a bare `Branch`. `createAndConnect` still requires a URI. The `branches.create` tool returns the same shape.
--- a/.changeset/sdk-branch-create-result.md
+++ b/.changeset/sdk-branch-create-result.md
@@ -1,0 +1,6 @@
+---
+"@neon/sdk": minor
+"@neon/tools": minor
+---
+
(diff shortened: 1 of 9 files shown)


andrelandgraf:
Closing. `branches.create` mapping to `Branch` is the [#493](https://github.com/neondatabase/neon-pkgs/pull/493) contract, not a mapper bug.

That release attached compute on create by default and split credentials onto `createAndConnect`. The 3.0.0 changelog: `create` returns the resource without a connection string; `createAndConnect` returns a URI. Current JSDoc on main still says this method keeps a branch-only result even when it provisions compute.

```ts
const { data: branch } = await neon.branches.create(projectId, { name: "preview" });
branch?.id;

const { data: connected } = await neon.branches.createAndConnect(projectId, {
  name: "preview",
});
connected?.connectionString;
```

`createAndConnect` is the URI job: `{ branch, endpoint, connectionString }`, `pooled`, and a client error if the response has no URI. Putting `connectionUris` on every `create` collapses that split, breaks `data.id`, and puts credentials on the default path. `projects.create` still returns `Project` the same way.

Endpoint metadata without a connection string can be a later change. Not this PR.

PR #595, opened 2026-09-10
fix(cli): handle disabled LFC in inspect

### LINKED ISSUE #589: neon inspect db: lfc-hit-rate and working-set return empty/misleading output on large fixed-size computes (LFC disabled)
## Summary

On large fixed-size computes (Neon CU ≥ 18), Neon now disables the Local File Cache (LFC) and sets `shared_buffers` to 75% of RAM (see [Improving Postgres compute cache on Neon, Part 1](https://neon.com/blog/improving-lakebase-compute-cache-part-1)). Two `neon inspect db` checks read the LFC directly, so on those computes they return empty or misleading output instead of reflecting the actual cache.

## Affected checks

In `packages/cli/src/utils/inspect_queries.ts`:

- **`lfc-hit-rate`** reads `neon_get_lfc_stats()` (`file_cache_hits` / `file_cache_misses`). With the LFC off, there are no stats, so it falls to `emptyMessage: "No LFC stats available."`
- **`working-set`** reads `current_setting('neon.file_cache_size_limit')` and `approximate_working_set_size_seconds()`, and reports `Exceeds Lfc` against the LFC size. With the LFC off, the LFC size is absent/zero and the comparison is meaningless.

Neither check knows about `shared_buffers`, so a user on a large fixed compute gets no usable cache diagnostics from these two checks even though the cache is working fine (served from shared buffers).

## Suggested fix

Detect the LFC-disabled state (e.g., `neon.file_cache_size_limit = 0`, or a large `shared_buffers`) and either:

1. Fall back to shared-buffers-based hit-rate / working-set metrics, or
2. Print a clear note like "LFC disabled on this compute; the cache is served from shared_buffers" and point to `SHOW shared_buffers` and the Compute cache hit rate metric.

## Docs

The Neon docs have been updated to note this caveat: website #5812 (compute-cache model) and #5816 (the `inspect` reference).

cc @sav-maya — you authored these queries; flagging for the CLI side. Happy to help with doc wording once the tool behavior is decided.

sav-maya:
## PR Checklist

- [x] Addresses an existing open issue: fixes #589
- [ ] That issue was marked as `status: accepting prs` (the issue is assigned directly)

## Overview

Detects when `neon.file_cache_size_limit` is absent or zero. In that case, `lfc-hit-rate` and `working-set` return their existing metric fields as `null`, add a separate `note`, show the active `shared_buffers` size, and direct users to the Compute cache hit rate metric. Null-only optional columns are hidden from table output, while JSON and YAML remain structured. Existing LFC-backed behavior is unchanged.

The disabled path short-circuits both Neon metric functions, so it does not attempt to read LFC statistics or calculate an LFC working set.

## Testing

- `pnpm --filter neon... build`
- `pnpm --filter neon exec vitest run src/commands/inspect.test.ts src/list_tables.test.ts` (57 passed)
- Biome check on changed TypeScript files
- Both SQL paths exercised against local Postgres; disabled-mode stubs throw if either Neon metric function is called

### DIFF
--- a/.changeset/inspect-disabled-lfc.md
+++ b/.changeset/inspect-disabled-lfc.md
@@ -1,0 +1,5 @@
+---
+"neon": patch
+---
+
+`neon inspect db lfc-hit-rate` and `working-set` now explain that the cache is served from `shared_buffers` when the Local File Cache is disabled.
--- a/packages/cli/src/commands/inspect.test.ts
+++ b/packages/cli/src/commands/inspect.test.ts
@@ -4,6 +4,7 @@ import { describe, expect } from "vitest";
 import { test } from "../test_utils/fixtures";
 import { INSPECT_QUERIES } from "../utils/inspect_queries.js";
 import { writer } from "../writer.js";
+import { visibleInspectFields } from "./inspect.js";
 
 // `inspect db` opens a real Postgres connection with the embedded wire client,
 // so these tests drive the hermetic `--db-url` path: point at a port nothing is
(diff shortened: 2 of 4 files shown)


andrelandgraf:
Andre's agent: Two gaps need to be closed before this has the required regression coverage.

1. The disabled-LFC path is not executed by committed tests. `inspect.test.ts` checks SQL substrings and passes fabricated rows to `visibleInspectFields`; the live inspect suite does not invoke `lfc-hit-rate` or `working-set`. That leaves the load-bearing production condition (`neon.file_cache_size_limit` is absent or zero on the affected fixed computes) and the SQL short-circuit unverified.

Possible coverage:
- Execute the exact catalog SQL against a Postgres fixture with the custom GUC set to zero and diagnostic functions that fail if evaluated. Assert both result rows, including null metrics and the note.
- Add a fixed-compute live smoke that records the actual GUC and runs both CLI commands.
- Assert table, JSON, and YAML output for enabled and disabled LFC.

2. `ratio` is now an all-null optional field. On enabled LFC before hit/miss counters exist, the query returns `ratio: null` and `note: null`; `visibleInspectFields` then reduces the table to `Name` only. That changes enabled-LFC behavior and gives no explanation.

Possible fixes:
- Keep `ratio` visible whenever the disabled-LFC note is absent.
- Model the zero-sample state separately and return an explanatory note for it.

Add the enabled/zero-counters case to the output tests whichever contract is chosen.

The SQL currently matches mcp-server-neon#349 semantically. The source-commit comment documents parity but cannot enforce it; a shared versioned query catalog is a reasonable follow-up, not a blocker for this fix.

